PR #11, opened 2019-05-29
Enforce common code-style via linter & travis-ci

JPSchellenberg:
Hey,

I also enforced a common code-style via linter & travis like in the H5P-Nodejs-library and converted everything to camelCase.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+src/client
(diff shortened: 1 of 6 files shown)


rtens (approved):
I propose to have one `ci` script that combines everything to be done on the CI server so it can be more easily reproduced locally, for example with a `pre-commit` hook that runs `npm run ci`

PR #17, opened 2019-06-05
getLibraryData: resolve editorDependencies within javascript field

JPSchellenberg:
Hey,

I noticed, that the getLibraryData method needs to resolve the editor-dependencies, which are found in a `library.json` under the `editorDependencies`-field and attach them under the `javascript`-field.

At the moment the tests fail and I can not figure out why.
@rtens do you see my mistake?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3427,6 +3427,11 @@
             "integrity": "sha1-8QdIy+dq+WS3yWyTxrzCivEgwIE=",
             "dev": true
         },
+        "h5p-nodejs-library": {
+            "version": "0.2.1",
+            "resolved": "https://registry.npmjs.org/h5p-nodejs-library/-/h5p-nodejs-library-0.2.1.tgz",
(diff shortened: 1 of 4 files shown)


JPSchellenberg:
I found the mistake:
I am using the `H5P-Nodejs-library`-`_loadAssets`-method to resolve the editor dependencies. While the `H5P-Nodejs-library` expects a sync function as `loadLibrary` the `H5P-Editor-Nodejs-library` expects an async function that returns a promise. We should make the implementation consistent. I would prefer the async way. I have no idea how hard it is to implement a recursive dependency-resolver with promises, do you @rtens?

rtens:
You're right, the function should be async. So did that [here](https://github.com/Lumieducation/H5P-Nodejs-library/pull/45). Can you make a new npm release?

JPSchellenberg:
I published the H5P-Nodejs-library with the async loadLibrary as v0.3.0 and implemented the newest version here. Everything seems to work.

PR #20, opened 2019-06-06
CI: Run npm install

JPSchellenberg:
Hey,

does it make sense to run `npm install` in our ci-script on travis-ci to catch install-errors?


rtens:
Travis already runs `npm install`. The tests would not pass without it.

PR #23, opened 2019-06-16
Content cache

sr258:
This pull request is a possible implementation of a working "ContentTypeCache" (see [Issue 5](https://github.com/Lumieducation/H5P-Editor-Nodejs-library/issues/5)). I'd like to discuss a few things before the pull request can be accepted, though:

1) While working on the ContentTypeCache several services that have not been implemented so far were required. I have implemented them (in a mockup or example implementation; you'll don't have to use them, but they were needed for the "cache" to work)
    - information about the current user (e.g. right to update content types) --> class User (in test/mockups)
    - enumeration of installed library (+ some more functionality) --> class FileLibraryManager (in test/mockups) (manages libraries as directories in the file system instead of using a database to store information about it; not complete yet)
    - Configuration settings --> class Config
    - Storage of simple key-value pairs --> class InMemoryStorage (in test/mockups)
   I think that it would be advisable to avoid having one big "storage" interface that aggregates various unconnected storage functions into one class, as the PHP implementation does. I think that the key-value library store should be a different class than the library storage, as they do totally different things. 

2. It has turned out that the ContentTypeCache does much more than just caching. That's why I've split it into two parts: ContentTypeCache and ContentTypeInformationRepository. You can check out what the later does in the inline documentation of the class.

3. I've made heavy use of the async & await keywords as it simplifies code. I hope this is fine for you.

4. Do you want to de-couple HTTP request to other servers (e.g. the H5P Hub) from specific HTTP libraries (such as axios, which I've used)? I don't think this is necessary, as axios allows testing with mockups, but what's your opinion here?

5. I have to implement more tests (particularly emulate HTTP responses from the Hub).


sr258:
I've added tests and now the "cache" should do the same as the PHP implementation does.

JPSchellenberg:
Hey,
thank you for your work. Unfortunately I was really busy this week, but I will have a look at it today and provide you with feedback.

JPSchellenberg:
I have to admit, you put in a lot of effort and I underestimated how complex the content type cache is. Thank you really much for it. :)

1. I would agree to not have one big storage interface. - I will try to make a list of interfaces that have to be implemented by the user.
2. Ok, that makes sense.
3. That's no problem. We will add a transpiler.
4. We want to have as few dependencies as possible, but I think there is no way around things such as axios, without putting a lot of work in it.

I would like to merge this PR as soon as possible and plan a roadmap/milestone for the remaining todos as issues. - Feel free to add issues as you like.

For some reasons travis still fails - I will have a look at it, as the tests seem to pass.

src/index.js
```diff
@@ -3,17 +3,20 @@ const H5P = require('h5p-nodejs-library');
...
 const defaultTranslation = require('./translations/en.json');
 const defaultRenderer = require('./renderers/default');
-const defaultContentTypeCache = require('./content_type_cache');
+
+const ContentTypeCache = require('../src/content-type-cache');
+const ContentTypeInformationRepository = require('../src/content-type-information-repository');
 
 class H5PEditor {
-    constructor(storage, baseUrl = '/h5p', ajaxPath = '/ajaxPath?action=') {
+    constructor(storage, baseUrl = '/h5p', ajaxPath = '/ajaxPath?action=', storage2, config, libraryManager, user) {
```
JPSchellenberg: I think we should rename `storage2` to something like `valueStorage` as `storage2` is not very descriptive.

JPSchellenberg (changes requested):

src/index.js
sr258: I've renamed it to keyValueStorage

JPSchellenberg (approved):

Issue #47, opened 2019-08-06
Dependency-order is wrong in some content-types


JPSchellenberg:
When I use certain content-types (course-presentation, interactive-video, etc..) , the depedency-order seems to be wrong. It looks a lot like the error from the renderer where the depedency-order was wrong, due to the async-libraryLoader: See [here](https://github.com/Lumieducation/H5P-Nodejs-library/issues/47).

![Screenshot 2019-08-06 at 21 56 52](https://user-images.githubusercontent.com/9697985/62572954-29a7a000-b895-11e9-9728-7b7af355d22f.png)

We have two options:
1. Use an sync-libraryLoader in the editor as well.
2. Get a async-libraryLoader working. (See Lumieducation/H5P-Nodejs-library#66)

What do you think? @rtens @sr258

rtens:
I'll integrate the async library loader from the viewer. Should solve this problem.

rtens:
So after an inital investigation it seems that the problem is not that the libraries are loaded in the wrong oder, but tha tthe libraries of the content are not loaded at all. At least not in the [render function](https://github.com/Lumieducation/H5P-Editor-Nodejs-library/blob/master/src/index.js#L39).

@JPSchellenberg Do you agree with my conclusion or did I overlook something?

JPSchellenberg:
@rtens I can not agree with your conclusion, as the libraries are loaded in my case:
<img width="853" alt="Screenshot 2019-08-08 at 14 55 06" src="https://user-images.githubusercontent.com/9697985/62704816-86699e80-b9ec-11e9-8b4e-c7ce64a46282.png">

Or which specific libraries do you mean?

Were the necessary libraries installed in your `/libraries`-folder?

JPSchellenberg:
On second thought: I think the libraries are loaded from the client-side-editor-code and not determined on server-side. But then we have to figure out how and where the order is determined.

JPSchellenberg:
I think the order is determined in the axjax call to `GET /ajax?action=libraries..`:

<img width="851" alt="Screenshot 2019-08-08 at 15 05 12" src="https://user-images.githubusercontent.com/9697985/62705583-ef054b00-b9ed-11e9-91e1-9332ed4bcf7b.png">

so we have to investigate the `getLibraryData`-method.

rtens:
Ah thanks that's exactly the pointer I needed.

Issue #12, opened 2019-05-31
Implement installLibrary


JPSchellenberg:

sr258:
I've looked into the code of the PHP implementation and installing libraries does the following:

1. Check token (see #25)
1. Check machine name (empty)
2. Check if machine name is part of hub (by checking in content type cache)
3. Check if user can install this specific content type
4. Download content type from Hub (by downloading from .../content-types/H5P.Example)
5. Validate package (H5P-PHP-library: H5PValidator->isValidPackage)
6. Store package (H5P-PHP-library: H5PStorage->savePackage)
7. Remove temporary files

The H5PStorage.savePackage function is also reused when uploading H5P packages with actual content, so implementing the installation will also advance uploading packages in general a lot. The whole process is pretty complex, though! What's next on your agenda?

The more and more I look at the H5P server core code, the more I get the impression that it's not a good idea to follow the structure of the PHP implementation! It seems to suffer from technical debt and isn't object oriented to a large extend (or mostly actually). Everything is in very few mega classes that do many unrelated things and the methods in those classes are sometimes very long and could be split up to make them easier to understand. 

Have you thought about how the architecture of the Node implementation should look like? Even though this might be difficult at the moment without having fully understood everything that the PHP implementation does, but I really think it would be good if there were a few sketches of the basic building blocks of the future application.

JPSchellenberg:
Ups. I have missed this comment. Sorry for answering so late. This one will be next on our agenda. The PHP implementation definitely suffers from technical debt and we should use our advantage to write code from scratch. I have not thought about the architecture as my approach till now was more like a 'try and error'-way by just replicating the server responses from the WordPress-plugin as I could not get my head around the php-code. (I don't know any php). And as I worked step by step I have no overview of what else the php-library does.

If you have any ideas I would be more than happy to hear them. Otherwise, we could have a brainstorm with @rtens .

JPSchellenberg:
@sr258 is there something we can do to help you with the validator?

sr258:
I‘ll integrate it into the editor today and create a pull request. It would be great if your reviewed it then. It has become pretty long (but the PHP validator is long and complex, too). You can check out the validator here: https://github.com/sr258/H5P-Extractor/tree/validator

sr258:
I've put the validator into the editor and create a branch for this: https://github.com/sr258/H5P-Editor-Nodejs-library/tree/hub-install
[...]

PR #48, opened 2019-08-06
Add puppeteer test (headless chrome test)

JPSchellenberg:
Hey,

today I found out, that the depedency-order seems to be faulty. This seems to be the same problem like the one with the async-libraryLoader in the H5P-Nodejs-library. I already opened an issue (#47) and this PR adds a test with puppeteer to see if the editor throws any error when accessing certain pages.

The problem is, that the test needs the libraries to be installed. Right now it loads the 'Library Install' page and then does nothing, so in order to get the test running, we need to implement the installLibrary-function (#12). I think @sr258 is working on the library-validator in order to install libraries. We can either help @sr258 with the validator or we implement the installLibrary without validation and add the validtor later on.
@rtens @sr258 How do you want to proceed?


rtens:
@sr258 seems to be making great [progress with the validation](#50), but I don't see a reason why not to go ahead and implement installLibrary without validation first. 

@sr258 are you working on validation only or also on installation?

sr258:
My next step would have been to work on the installation process of libraries. The validator will be finished soon, so it would make sense to continue there. I‘m on holiday at the moment, so I will have some time to work on it. 

I thought of extending FileLibraryManager which is currently a mock-up by separating it into a generic LibraryManager class and a FileLibraryStorage class that does the actual storage. That way storage can also be in a Redis Cache or on a S3 by exchanging FileLibraryStorage with RedisLibraryStorage etc.

rtens:
Nice. Sounds good to me =)

rtens:
@JPSchellenberg Do you still wanna merge this? If so, then go ahead. Otherwise let's close this PR.

JPSchellenberg:
@rtens I am still working on it as the tests do not pass, but I plan on fixing that and merging this today as it depended on the installLibrary-method.

sr258:
How about using Jest for the Puppeteer tests as well? It seems to work fine and this would allow better error reporting if there are more Puppeteer tests in the future: https://jestjs.io/docs/en/puppeteer

JPSchellenberg:
I will close this PR as it is easier to make a new branch from master and add the tests there.

PR #50, opened 2019-08-08
Validation of H5P Packages and translation service

sr258:
The PR adds a component that does validation of H5P packages and a translation service for strings. Currently, it is not referenced anywhere in real code but only in the tests. The next step would be to add tests for the real-life examples provided at H5P.org.

I had to move from babel-cli to @babel/cli because babel-cli is unmaintained and the build process didn't work with it because the old babel used by babel-cli didn't work with the @babel/plugin-proposal-class-properties plugin. I also upgraded other packages to their most recent version in that process.


rtens:
I'm a bit worried about the size of this branch. Any way we can integrate an interim state into master?

sr258:
All right. I've merged the current master.

Here is an overview of what has happened in this branch:
* adds the class properties proposal to babel
* adds automatic download of all content types (.h5p files) from Hub for testing purposes (done on install)
* extends config class with values necessary for the validator
* adds a basic install method to ContentTypeInformationRepository (this will be extended soon) and the required dependencies
* separated FileLibraryManager into LibraryManager (generic utilities) and FileLibraryStorage (specific implementation of a library repository)
* adds PackageValidator (does the validation of H5P packages in memory) and helping classes (ValidatorBuilder, ValidationError) + JSON schemas
* adds a TranslationService + English translations string for errors occuring during validation
* adds tests for all of the above + test data

The branch did indeed become very large, but validation isn't a piece of cake.

JPSchellenberg:
Wow. That is really a lot... thank you so much for your work as this seems to be quite complex and complicated. I think we are ready to merge, right?

rtens:
Looks pretty nice. Really not a piece of cake. Thanks for your work on that, @sr258 =)

PR #56, opened 2019-08-14
Fixed bugs in Express example

sr258:
This PR fixes three issues with the Express example:

1. https://github.com/Lumieducation/H5P-Editor-Nodejs-library/pull/54 introduced a bug because saveContentFile existed twice with different parameters. Because of this uploading images was no longer possible.
2. Fixed code issues ESLint complained about.
3. Added example folder to dev-dependencies so that ESLint doesn't complain about the imports.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -24,6 +24,7 @@
             "LabeledStatement",
             "WithStatement"
         ],
-        "no-param-reassign": "off"
+        "no-param-reassign": "off",
+        "import/no-extraneous-dependencies": ["error", {"devDependencies": ["test/**.js", "examples/**.js"]}]
     }
 }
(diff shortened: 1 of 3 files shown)


rtens (changes requested):
Ah sorry for overlooking that there was already a function for that. Could you merge the two into one? Either by turning the Buffer into a Stream or vice versa.

Also, what is the `field` argument meant to be used for?

sr258:
Merging isn't as easy it appears without changing a bit more. There must be some kind of ContentManager that managers storing content and access to it. I think it makes sense to think of a new interface when the ContentManager is designed and leave it as it is now. It's a hack anyway and it gets things done for the time being. The other use of saveContentFile(...) will have to be superseded by proper management of temporary files in a TemporaryFileManager or something like this.

The ``field`` property is the property of the content.json ("params" in H5P jargon) that an image is added to. It is not needed for storing a content file to the database.

rtens:
Alrighty let's leave it as it is for now =)

rtens (approved):

Issue #63, opened 2019-08-16
Documentation misses npm run build


jankapunkt:
The `npm run build` command is missing, which is required to run without errors.

sr258:
Fixed in PR, so the issue can be closed.

Issue #62, opened 2019-08-16
Configurable Config


jankapunkt:
I need to configure a custom Hub and other options int the [config](https://github.com/Lumieducation/H5P-Editor-Nodejs-library/blob/master/src/config.js) but currently it's hard-coded.

sr258:
Implemented in PR, so the issue can be closed.

PR #64, opened 2019-08-17
#62 and #63: Configurable config and build command in readme

sr258:
The changes add a config.json file to the example folder in which the configuration settings can be changed without going into the code (#62). It also adds the ``npm run build`` command to the documentation. (#63)


sr258:
I've updated the branch. It should now be on level with the master and can me merged.

Issue #67, opened 2019-08-22
Downloading test data from H5P Hub fails at "Twitter User Feed" content type


sr258:
Downloading the "Twitter User Feed" content types produces a 404 error. It was probably removed.

JPSchellenberg:
Yep. I think it was removed.

PR #68, opened 2019-08-22
[#67] Updated Hub mockup data and added script to automate this update

sr258:
The automated example download broke because the Hub info we used was out of date. I wrote a script that downloads the Hub info for future use.


rtens (changes requested):
Please merge master and update the branch, since I removed babel and the "build" directory.

I only now noticed that this PR is open. Please request reviews from @JPSchellenberg and me in the future so we are notified.

sr258:
I've merged the master into the branch, so everything should be fine now. Sorry, I thought you received automatic notifications on this. Is there a role you can assign me, so I can assign reviewers?

rtens:
Ah I wasn't aware that you cannot request reviewers. We'll fix that.

rtens (approved):
I'll add the generated file to .gitignore since I assume it should not be version controlled.

sr258:
I've reverted the change in .gitignore. It makes sense to have it version controlled as the tests depend on it. We just have to update the file from time to time as the hub is updated. I don't want to have a live download of the file with every install, though, as it might break tests, when Joubel changes hub info.

rtens:
Ah sorry I misunderstood your intentions with this file. I'm generally suspicious of manual processes and try to avoid them. The solution is good enough for now but I hope we find a better one which doesn't depend on Joubel.

PR #78, opened 2019-08-27
[closes #76] revised dependencies

### LINKED ISSUE #76: Revision of dependencies
Dependencies should be checked for the following criteria:

* is it a dev-dependency, which accidentally has been installed as dependency?
* is it a hard-required dependency (for example to run the editor)
* is it an optional dependency?
* ... what else?

Example: `fs-extra` is currently in dependencies but i only used as dev-dependency.

rtens:
See #76 

I removed some not needed dependencies. Most importantly, I removed babel which means we have to stick with JavaScript as implemented by Node v10.16 which I think is a good thing since it makes the project leaner.


sr258 (commented):
Everything fine from me. We'll have to think about a good eslint ruleset for the second revision, though.

PR #99, opened 2019-09-17
Add getLibraryFileUrl to library-manager (closes #97)

### LINKED ISSUE #97: _libraryManager.getLibraryFileUrl is not a function
The [getLibraryFileUrl](https://github.com/Lumieducation/H5P-Editor-Nodejs-library/blob/master/src/content-type-information-repository.js#L135) function for the libraryManager is not defined.

JPSchellenberg:
Hey,

I noticed that the getLibraryFileUrl-method of the libraryManager does not exist. It is used within the [content-type-information-repository](https://github.com/Lumieducation/H5P-Editor-Nodejs-library/blob/master/src/content-type-information-repository.js#L135) to determine the URL of the icon.svg.

### DIFF
--- a/src/library-manager.js
+++ b/src/library-manager.js
@@ -200,6 +200,20 @@ class LibraryManager {
     }
 
     /**
+     * Returns an URL where the requested file can be accessed.
+     * @param {Library} library 
+     * @param {string} filename
+     * @returns {string} the url
+     */
+    getLibraryFileUrl(library, filename) {
+        try {
+            return `${library.machineName}-${library.majorVersion}.${library.minorVersion}/${filename}`;
+        } catch(error) {
+            return undefined;
+        }
+    }
+
+    /**
      * Updates the library to a new version. 
      * REMOVES THE LIBRARY IF THERE IS AN ERROR!!!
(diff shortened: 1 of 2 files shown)


sr258:
I erroneously removed gitLibraryFireUrl from FileLibraryStorage as I thought that it's not needed after all, but it looks like I was wrong and we need it after all. I'm not sure about putting it in LibraryManager, though. 

Resolving the library and filename combination to a URL should be done by the component that also serves the files - this would be the express implementation in the demo application. The FileLibraryManager doesn't have to know how and where its files are served, so I wouldn't put it here. 

How about passing the resolution function to the constructor of LibraryManager? That way the components can access the function when its needed and its still up to the implementation how and where it wants to serve the files?

JPSchellenberg:
The constructor of the LibraryManager is not called by the user, but by the H5PEditor constructor. So if we want the user to implement this function, we would have to add it to the Editor-constructor.

sr258:
You're right. I opt for passing it to the editor constructor then. @rtens ?

JPSchellenberg:
I will close this PR and open another one for the new Typescript one.

PR #100, opened 2019-09-17
listLanguages: return empty array if language folder does not exist (closes #98)

### LINKED ISSUE #98: file-library-storage: getLanguageFiles should return empty array if no languages are found
The [getLanguageFiles](https://github.com/Lumieducation/H5P-Editor-Nodejs-library/blob/master/src/file-library-storage.js#L69)-method should return an empty array if it fails to load the languages.

JPSchellenberg:
The libraryManager listLanguages-method should return an empty array if no langauge-folder can be found within the library.

### DIFF
--- a/src/library-manager.js
+++ b/src/library-manager.js
@@ -161,7 +161,12 @@ class LibraryManager {
      * @returns {Promise<string[]>} the language codes for translations of this library
      */
     async listLanguages(library) {
-        return this._libraryStorage.getLanguageFiles(library).catch(error => [])
+        try {
+            return await this._libraryStorage.getLanguageFiles(library);
+        }
+        catch (error) {
+            return [];
+        }
     }
 
     /**
--- a/src/library-manager.js
+++ b/src/library-manager.js
@@ -161,8 +161,7 @@ class LibraryManager {
      * @returns {Promise<string[]>} the language codes for translations of this library
      */
     async listLanguages(library) {
-        return this._libraryStorage.getLanguageFiles(library);
-
+        return this._libraryStorage.getLanguageFiles(library).catch(error => [])
     }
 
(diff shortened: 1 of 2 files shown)


src/library-manager.js
```diff
@@ -161,8 +161,7 @@ class LibraryManager {
      * @returns {Promise<string[]>} the language codes for translations of this library
      */
     async listLanguages(library) {
-        return this._libraryStorage.getLanguageFiles(library);
-
+        return this._libraryStorage.getLanguageFiles(library).catch(error => [])
```
sr258: I think we should settle on a style to handle promises for the project:

I prefer the async await style as it allows you to write code that look like synchronous code and you don't have lots of nested callbacks. If you don't use async await the changing identation often also confuses change comparison tools and it becomes more difficult to see what has actually been changed and what is simply new indentation because a function call has been added or removed somewhere. This was the case when merging in changes in index.js earlier in the project, and created the need for some longer manual comparison of changes.

sr258 (dismissed):
All fine, just a remark on style consistency.

src/library-manager.js
JPSchellenberg: I am not too familiar with the async await style. How would that look like? Wrap it in a try catch block?

sr258: In this case it would be:

```js
try {
  return await this._libraryStorage.getLanguageFiles(library);
{
catch {
  return [];
}
```

As it‘s only a one liner the benefit isn‘t that obvious, but in longer functions it becomes very visible.

JPSchellenberg: Alright. I see the advantage. I will change it as soon as possible.

sr258 (approved):

PR #101, opened 2019-09-17
Add prettier to enforce common code-style

JPSchellenberg:
Hey,

today I noticed that the Editor misses a format check via prettier and I would strongly recommend to run it to enforce a common code style & reduce format noise. The player already has this format check. You can install a prettier plugin in your IDE which automatically formats while working.


sr258 (approved):

sr258:
We could also re-add a ESLint configuration to catch other code style issues. I wasn't too keen on the AirBNB style we had some time ago, but I'm open to any convention. (Not sure why it was removed.)

PR #103, opened 2019-09-18
Move to Typescript

JPSchellenberg:
Hey,

As far as I remember we all agreed that moving to Typescript would be a good thing for the project. I would like to Typescript as soon as possible, because it makes the development process easier.

I made a small todo-list. Feel free to add Todos and work on this branch. If you are unsure about changes, just make another branch and create a PR for this branch.

Todos:
- [x] Define Interfaces, classes & types
- [x] Convert every file (src & tests) to Typescript
- [x] Get TSLint running
- [x] Get Jest running with Typescript


sr258:
A few things I've come across:

- Shouldn't we also remove types from JSDoc? They are already defined in TS, so the JSDoc types are redundant now?
- I think it would be good if we ordered all class methods alphabetically and enforce it through a TSLint rule. This simplifies merging in case of merge conflicts.
- Should all files be rename to UpperCamelCase / PascalCase like in ContentManger.ts?
- Should we rename all private properties named _propertyName to propertyName?
- I'd also suggest removing "variable-declaration" from the typedef TSLint rule. I think implicitly setting variables tot types is ok.
- We should also move remove direct references to the classes User and EditorConfig in favour of referencing the interfaces IEditorConfig (must be created) and IUser. That way the implementation of the interfaces can be exchanged by the implementation/plugin.

JPSchellenberg:
1. I wouldn't remove the JSDoc annotations as they pop up in my IDE (vscode) and provide additional information, which I find very useful: 
Screenshot 2019-09-21 at 17 12 39

2. I agree. I will look for a corresponding TSLint rule.

3. I am not sure about this one. I find it useful when the file is called exactly like the exporting class. What do you think?

4. I am also not sure about this one. What do you think? I thought it is best practice to just prefix them with the `private` marker.

5. Ok.

6. You are right.

JPSchellenberg:
We should also coordinate who is working on which file. I am currently working on the LibraryManager.

sr258:
1. I use VS Code too, and this is what I get in a method without JsDoc type:
image
This is the same as in your screenshot and we avoid having a second place where the types are specified (which might not be changed when the method signature is changed...)
2. -
3. Let's use filenames like "ContentType.ts" then. It also makes TSLint happy...
4. Yes, I would remove the underscore. There is no need for convention, as private members can't be accessed from outside anyway.

sr258:
I've removed the need to explicitly declare types in arrow functions. I think this is not necessary as arrows functions are mostly used in lightweight situations, where typing becomes very verbose. 

The validator's tests succeeds now, but I must say that the way the PackageValidator class looks now is not prettier's most impressive hour...

JPSchellenberg:
Well, I got everything running except the tests in `test/getLibraryData.test.js`. I think they mostly test things that are covered in Typescript now. For example, one test defines how the semantics-object should look like, but we already test this with Typescript. @sr258 could you have a look and tell me what do you think? Should we just throw away these tests or are there some that might be helpful?

sr258:
I think many of the tests in getLibraryData are still useful even with TS:

- includes assets of preloaded and editor dependencies
- includes assets of preloaded and editor dependencies
- loads the language
- lists all available languages
- lists all available languages

They actually do test logic inside the system, so I think we should keep them.

sr258:
Have you got proper debugging working for TS in VS Code? In general it does work for me, but when I use breakpoints and step-by-step debugging the cursor jumps all over the files. This seems to be a problem with the source-maps generated by TS.

There are reports that this is connected to the coverage generation (e.g. https://github.com/kulshekhar/ts-jest/issues/917), but turning it off doesn't help. Using 2.3.1 of ts-jest, as suggested in the thread, didn't help. There are more issues like this around: https://github.com/kulshekhar/ts-jest/issues/46

JPSchellenberg:
I have this set up in `launch.json`: https://gist.github.com/JPSchellenberg/5192a15ab5ce9c4f94adf37c57c1640a and it seems to work decently. 
Ok, I will work on the getLibraryData tests and convert them.

JPSchellenberg:
Now everything compiles successfully, the code is linted and formatted and all tests pass. I think we are good to merge or did I forget something?

sr258 (approved):
I think everything is in order. It's worth thinking about what to do with types.ts, though. We should identify interfaces that are necessary for the implementation to know about and put those into a special file. Other internal interfaces could be put somewhere else. I'm not sure what the best practice for interface files in TS is, but I have the feeling that we shouldn't put all interfaces into one file.

PR #105, opened 2019-09-24
Replace Travis with  CircleCI

JPSchellenberg:
Hey,

I figured out that CircleCI allows you to configure multiple stages in your ci, which show up in GitHubs PRs: 
Screenshot 2019-09-24 at 13 36 31
I find it useful to see in the PR at which stages the ci fails and would recommend to switch to circleci. What do you think?


sr258 (approved):
Looks more useful than just a "pass/fail".

PR #109, opened 2019-09-25
Run download-script only when installed via npm isntall (fixes #107)

### LINKED ISSUE #107: Postinstall scripts should not be run when installed in another library
Currently, the postinstall script calls:

```npm run download:content-type-cache && npm run download:content```

This call is also executed when the h5p-editor-nodejs-library is installed in another package, e.g. in h5p-demo. As the files downloaded here are only meant for development purposes, they should not be downloaded in this situation.

JPSchellenberg:
Hey,

as @sr258 mentioned in issue #107 the postinstall script also runs when installed as dependency. We had the same issue in the Player and fixed it by putting these script into the `prepare` field rather than `postinstall` field. The `prepare` scripts only run when used via `npm install`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     },
     "scripts": {
         "build": "./node_modules/.bin/tsc",
-        "postinstall": "npm run download:content-type-cache && npm run download:content",
+        "prepare": "npm run download:content-type-cache && npm run download:content",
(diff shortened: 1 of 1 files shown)


sr258 (approved):
A test has shown that it works now as wanted.

PR #115, opened 2019-09-26
Add CONTRIBUTING.md

JPSchellenberg:
Hey,

I think we should provide some guidelines on how to open PRs and contribute to this project. I adapted the CONTRIBUTING.md from WordPress Calypso: https://github.com/Automattic/wp-calypso/blob/master/.github/CONTRIBUTING.md
Furthermore I think we should really strive for short-lived branches and small PRs as it makes the reviews easier.

 What do you think?


sr258 (approved):
I think the guidelines are very useful.

sr258:
One question: Should we merge PRs with failing tests (e.g. because a features is not complete yet, but full implementation takes time)?

JPSchellenberg:
No, all tests have to pass in order to be merged. The master branch should always contain working code. We have to hide unfinished features behind a feature-switch of something like that or find another solution.

PR #116, opened 2019-09-26
Merge H5P-Player-Nodejs-Library (closes #55)

### LINKED ISSUE #55: Merge h5p-editor-nodejs-library and h5p-nodejs-library
The more I look at h5p-editor-php-library and h5p-php-library, the less does the distinction between the two make actual sense to me: Many things that are clearly part of the editor (content and package validation, file upload, ...) are part of h5p-php-library. As far as I know the "player" can't be used without also having the editor in current PHP implementations. There are also many dependencies of the editor on the other library, which makes the whole h5p project difficult to understand for new programmers, because you constantly have to switch between projects and lose things like "Go to definition" in IDEs. 

So why have to separate projects at all? If there should be a standalone "player" application, you could always start the application with a special flag or change something in the configuration to disable the editor.

JPSchellenberg:
Hey,

as discussed in #55, this PR merges the H5P-Player into this repository.


JPSchellenberg:
I'll close this PR and create a new one for the player and a separate one for the e2e tests.

PR #118, opened 2019-09-26
Content export functionality (closes #113)

### LINKED ISSUE #113: Exporting content
We need to implement an exporter method that creates .h5p packages that can be downloaded by the user. 

I will implement the exporter in the editor as it will be merged with the player, and there are dependencies that we need for it in the editor.

Current work is in the branch export-content

sr258:
The branch adds content exporting functionality (= saving .h5p packages for content).


sr258:
Exporting works now, but there is currently no UI for it in the demo.

The tests import and export the content from the h5p hub to make sure that the exporter works the same way the php implementation does it.

What I've noticed is that the exporter is rather slow if you export huge packages like for CoursePresentation (includes many dependent libraries). I think this issue must not be addressed right away and that the priority should be on getting a functionally working version, soon.

sr258:
BTW: I also fixed some problems with the demo not working due to incorrect references.

JPSchellenberg (approved):

PR #120, opened 2019-09-26
Player: Add renderHtmlPage test

JPSchellenberg:
Hey,

I tried to add the [renderHtmlPage-test](https://github.com/Lumieducation/H5P-Player-Nodejs-library/blob/master/test/render-html-page.test.js) from the H5P-Player-Nodejs-Library, but it fails when comparing the generated HTML to a given string.

I tried to use `.replace(/\r?\n|\r|\s/g, '')` on the generated html and the string I am comparing to to get rid of the spaces and line endings, but it did not work.

Has anyone an idea on how to approach this correct?

### DIFF
--- a/test/H5PPlayer.renderHtmlPage.test.ts
+++ b/test/H5PPlayer.renderHtmlPage.test.ts
@@ -1,0 +1,288 @@
+import H5PPlayer from '../src/H5PPlayer';
+
+describe('Rendering the HTML page', () => {
+    it('uses default renderer and integration values', () => {
+        const contentId = 'foo';
+        const contentObject = {
+            my: 'content'
+        };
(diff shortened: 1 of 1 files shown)


sr258:
The reason for the failing tests is that the JSON objects included in the html code (H5PIntegration) is not in the same order in the actual result and the target string.

The actual values are ordered alphabetically (as might be expected due to our TS lint rule) but the strings they are checked against have a different order. So rearranging the order of the expected strings should work.

Normally you could use expect(...).toMatchObject(...) for comparing objects (the order doesn't matter then), but it's still not recursive...

JPSchellenberg:
Thank you! I haven't thought about that. I will change that asap.

PR #147, opened 2019-10-01
Add e2e-test for player

JPSchellenberg:
Hey,

I reworked the e2e-test via puppeteer we had in the H5P-Demo repository to run here and integrated it in the CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,6 +64,19 @@ jobs:
                   paths:
                       - ./node_modules
             - run: npm run test
+    player-e2e-test:
+        docker:
+            - image: circleci/node:10-browsers-legacy
+        steps:
+            - checkout
+            - restore_cache: # special step to restore the dependency cache
+                  key: dependency-cache-{{ checksum "package.json" }}
+            - run: npm ci
+            - save_cache: # special step to save the dependency cache
+                  key: dependency-cache-{{ checksum "package.json" }}
+                  paths:
+                      - ./node_modules
+            - run: npm run test:e2e:player
 workflows:
     version: 2
     h5peditor:
@@ -75,3 +88,4 @@ workflows:
             - lint
             - format
             - unit-tests
+            - player-e2e-test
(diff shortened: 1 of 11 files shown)


JPSchellenberg:
The test fails at the CoursePresentation type, but I can not replicate the error with the example.
Any ideas why this could fail?
Screenshot 2019-10-02 at 00 06 56

sr258:
First, I had the suspicion that the error is related to the fact that we are currently using version 1.22 of the H5P core and the CoursePresentation requires 1.23. But I tried updating to 1.23.1 [in this branch](https://github.com/Lumieducation/H5P-Nodejs-library/tree/try/update-1.23.1) and it didn't help (there's a new file in the core, so we can use the branch as a starting point for updating later).

What's strange is that CoursePresentation doesn't even include the InteractiveVideo but only a regular video! Is it possible that adding a regular video is a sub-function of InteractiveVideo?

My current hypothesis is that the play property is the property of the video object that should be played. The video object seems to be undefined. The reason could be that content objects don't seem to work in the E2E tests (images are not displayed). Maybe the error is thrown when H5P tries to load a video that's not accessible?

sr258:
I have a solution for the problem (for the time being?):

The reason of the test failure was that the H5PIntegration.saveFreq was set to '0' to disable saving user state, but must be set to 'false' for this. Setting it to 0 means that the h5p core tries to load the user state, which seems to fail: https://github.com/h5p/h5p-php-library/blob/117668d0e54305c5be93a891cbc75b9d61a12dab/js/h5p.js#L2810 

I've changed the value of saveFreq to 'false' to disable this. Now the tests succeed. I'm not sure what will happen if we finally implement saving user state and the line will be called again, but I think we can deal with this when we come to this.

(The force push is there because I accidentally pushed the update to version 1.23.1 of the core as well.)

JPSchellenberg:
That is great! I haven't thought about that... It is amazing how such tiny details matter.

Issue #154, opened 2019-10-04
Separate implementations from (core)-Editor


JPSchellenberg:
After thinking about the code for a bit I noticed that some of our files in `src/` are basically implementations:
- FileContentStorage.ts
- EditorConfig.ts
- FileLibraryStorage.ts
- InMemoryStorage.ts
- JsonStorage.ts
- User.ts

I think we should keep them in the Repo as examples and some kind of default implementations, but separate them from the source-code in `src/` as it makes it clearer how to implement the library.

@sr258 What do you think?

sr258:
Yes, I'm fine with moving them to some "implementation" folder or something like this. We should definitely keep them at hand, though, as they are needed by many tests.

PR #156, opened 2019-10-04
Move implementation files from src to examples

JPSchellenberg:
Hey,

as discussed in #154 I am moving the implementations from the `src/` folder to the `examples/` folder. 

I noticed that a lot of classes require the EditorConfig and I am currently working on replacing that with an IEditorConfig-interface.


sr258:
The application can't be started at the moment, if you transpile everything from a clean state. The reason is that index.js still expects the moved files to be in the old location. I think they shouldn't be included in index.js anymore at all as they are part of the example now. 

There is another problem: Currently, the files in the examples folder aren't transpiled. We'll have to run tsc for the example dir as well (and maybe move example.js to typescript, too?)

Another thing I've noticed is this linter error:
image

On the one hand, this is strange, as the imports ARE ordered alphabetically. On the other hand, this also shows that the linter does currently not work for the test folder. I think this is because we have this line in tsconfig.json: ```"include": ["src/**/*"]``` This means that tsc ignores the files outside the src folder. This is good, as we don't want to transpile the test. But this is also bad, because it means that the linter won't work for directories other than 'src'. (And VS Code, "find all references" also only works within the src folder). We should think of a way to fix this.

sr258 (changes requested):
The application can't be started at the moment, if you transpile everything from a clean state. The reason is that index.js still expects the moved files to be in the old location. I think they shouldn't be included in index.js anymore at all as they are part of the example now.

There is another problem: Currently, the files in the examples folder aren't transpiled. We'll have to run tsc for the example dir as well (and maybe move example.js to typescript, too?)

Another thing I've noticed is this linter error (ContentTypeInformationRepository.test.ts):
image
image


On the one hand, this is strange, as the imports ARE ordered alphabetically. On the other hand, this also shows that the linter does currently not work for the test folder. I think this is because we have this line in tsconfig.json: "include": ["src/**/*"] This means that tsc ignores the files outside the src folder. This is good, as we don't want to transpile the test. But this is also bad, because it means that the linter won't work for directories other than 'src'. (And VS Code, "find all references" also only works within the src folder). We should think of a way to fix this.

sr258:
- I've thought about the ordered-imports error from above and I think the best solution is to introduce a new import group for modules in the examples folders. 
- The problem with the references to TS files in test has been fixed in PR #162 . I think merging the commit there into this branch would solve the issues.
- If the commit from PR #162 is merged in, this would also be a good base to transpile the ts files in the ```examples``` folder. We have to add them the tsconfig.build.json file and think of a good place to store them so that all references work.

sr258:
I've added transpiling for the example implementations and removed references in the index.js file.

sr258 (approved):

PR #162, opened 2019-10-07
Typescript references now also work in tests

sr258:
This PR solves the problem mentioned in https://github.com/Lumieducation/H5P-Nodejs-library/pull/156 that TS references don't work across the ```test``` and ```src``` directory. Now IDEs will resolve references across these directories (for ts files, references in ```express.js``` can't be resolved):

- src
- test
- examples

The approach I took was the one explained in [the accepted answer here](https://stackoverflow.com/questions/35470511/setting-up-tsconfig-with-spec-test-folder): I've split up the old tsconfig.files into a base file that includes all common settings, one file that the IDE uses (tsconfig.json) and one that the npm build script uses (tsconfig.build.json). The tsconfig.json has its own ```outDir```, which is used in case somebody runs tsc directly.

Now TSLint also works in the test directory!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ npm-debug.log*
 
 # Compiled files
 build
+build-ide
 h5p
 coverage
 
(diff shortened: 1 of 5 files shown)


sr258:
I've just seen that the linter fails now. I don't want to fix these issues while you're working on restructuring the  project, @JPSchellenberg . Maybe you can merge this commit into your branch and fix the issues there?

PR #173, opened 2019-10-12
feat(log): Add logging module (closes #153)

### LINKED ISSUE #153: Add Debug Logger
Hey,

I am badly missing a debugging logger like [debug](https://www.npmjs.com/package/debug). Sometimes I get questions in Slack from people that want to integrate the library but fail to do so and without an error/info log it is really hard to see why it fails. We talked in #83 about a logger like winston, but that was more a solution to report errors back to people who are using the Editor via the web interface - as far as I understood it.

I am talking about a logger for developers that are implementing the library.

I would opt for [debug](https://www.npmjs.com/package/debug) as I am used to it and it is also used by a lot of packages like express.

@sr258 Do you have any other suggestions or comments?

JPSchellenberg:
Hey,

I started to add a logging-module. Currently it comes with three different log-levels: info, warn and error and it is based on [debug](https://www.npmjs.com/package/debug). 

I am not 100% sure where to put logs and where it gets to verbose, but I think this is just something we have to figure out with time or does anyone know any best practices?

Modules that use the logger:
- [x] ContentManager
- [x] ContentTypeCache
- [x] ContentTypeInformationRepository
- [x] DependencyGetter
- [x] H5PEditor
- [x] H5PPlayer
- [x] PackageExporter
- [x] LibraryManager
- [x] PackageImporter
- [x] PackageValidator
- [x] TranslationService


sr258:
I wonder whether we should generally log the „context“ a log call is executed in. With this I mean what user is doing something, what contentId is being worked on etc.

As my day-job involved lots of GDPR concerns for (state) institutions I think we should make sure that admins can configure what user information is logged! If we want the library to be used by education facilities in Europe, they will need the ability to control what personal data is stored in the logs (just logging what a user did is personal information). At the same time non-personal logging should still work.

So I propose that we always add the user object to the log and a „log message id“ or something similar. The logger can then determine whether the log entry should be stored or discarded according to privacy rules (e.g. do not log student or teacher activities, but do log admin activity). Of course, this would mean having lots of „log message ids“, which might be annoying to manage...

JPSchellenberg:
My intention and main goal with the logger is to improve the development and feedback process, so user-information is not important at all. So I would propose to remove all user-information and just log development-relevant information like if a package did not validate and where it failed.

I think if someone wants to see what a user is doing, he should implement that by himself.

If we want more sophisticated logging, we could add that later, but for now, I want users to be able to paste some logs if they open an issue or contact me via slack.

sr258:
Alright, I see the usefulness of debug logging.

sr258:
I've look at the code in the branch and it looks fine, but I don't know how to actually enable the logger in my dev environment. Maybe we should add this to the documentation.

I have set this in launch.json:

```
"env": {
                "LOG_LEVEL": "DEBUG",
                "DEBUG": "*"
            }
```
I now get debug messages from express, but not from our modules.

JPSchellenberg:
You need to set the DEBUG_LEVEL to „debug“ (lowercase). That should fix it, but maybe we should make this case insensitive.
We should definitely add this to our documentation. I am currently working on the documentation.

sr258 (approved):
Got it now. The problem seems to be with my VS Code installation. If I simply add ```DEBUG=* ``` in the command line it works. Looks like the Windows VS Code version has problems with setting environment variables for debug profiles...

For future reference: There is a working launch.json file [here](https://github.com/Lumieducation/H5P-Nodejs-library/wiki/sr258%27s-launch.json-file-for-VS-Code).

PR #174, opened 2019-10-12
ci(release): Add semantic-release for automated releases & version-mgmt

JPSchellenberg:
Hey,

this PR will add the [semantic-release](https://github.com/semantic-release/semantic-release) module, which manages releases and versions using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) (See also #172 ), so we don't have to worry about releasing new versions.


sr258 (commented):
Looks like a useful addition. I have a question about when releases are generated, though: If we auto-generate releases on master, won‘t we have way too many major and minor release with very high version numbers? Every commit with a new feature would mean we have a new minor release and every breaking change a new major release. At the current stage of the project this would happen very often. I think we should run the bot on a special release branch to avoid this. Or am I misunderstanding how semantic-release works?

JPSchellenberg:
No, you understood it right. It would release a new minor version with every feature and a new patch with every fix, but that is actually supposed to happen. We could end with high version-numbers, but we always ship the latest version. See [here](https://semantic-release.gitbook.io/semantic-release/support/faq#is-it-really-a-good-idea-to-release-on-every-push). I think the biggest advantage is, that it forces us to follow best practices and not change the interface too often.
But I also agree that we might change the interface a few more times before we are comfortable. But as soon as we are comfortable we should name it v1.0.0 and stick to the semantic-release pattern. So I would suggest merging this PR as soon as we hit v1.0.0, but till then we put it on hold.

JPSchellenberg:
I will merge this PR, as I think it makes more sense to frequently release new versions.

Issue #146, opened 2019-10-01
Delete content


sr258:
There should be operations that allow users to delete content.

Issue #112, opened 2019-09-25
Place to check access rights


sr258:
Where do we want to check if a user has the permission to perform an operation. For example, where do we check if a user has the permission to download a piece of content. Is this done in the exporting function or is the exporting function unaware who is calling it and the permissions check are done e.g. by the implementation?

JPSchellenberg:
I would strongly advocate for letting user permission up to the implementation. I don't see a reason to do it in the library or am I missing something?

sr258:
Well, who has which permissions must obviously come from the implementation, but the library could check if users are actually allowed to do an action on its own, e.g. like here: https://github.com/Lumieducation/H5P-Editor-Nodejs-library/blob/a90d8de33413845ae0b56cd09fd462e5458a7a90/src/PackageManager.ts#L120

It might be possible to completely keep the library free from permissions and fully handle them before the library is called, but this would mean that the implementation has to call very specific methods in some cases. Consider this second example:

When a user uploads a package, the library must know whether the user has privileges to install libraries. One way of approaching this would be to simply call the library method and the library inquires which privileges the user has and acts accordingly (installs libraries or checks if the content can be used without installing it). An alternative would be to leave the check to the implementation and provide to methods: uploadPackageAndInstallLibraries(...) and uploadPackageWithoutInstallingLibraries(...).

After writing this, I think the library should be able to check what permissions a user has to not overcomplicate things for the implementation.

jankapunkt:
I would really like to continue to check permissions on the implementation level. For me it would be more complicated if  I have also consider module-specific permissions in addition to those within my application.

Issue #83, opened 2019-08-30
Reporting back status with proper AJAX responses


sr258:
(This has also been discussed in #43. #80 is also related, as all currently unlocalizable error messages in the code need to be replaced if we've decided how to manage reporting errors)

The H5P client expects errors (and more detailed success messages) to be a part of the AJAX response in several cases.

Examples:

- it gets a (success) message about how many libraries were added or updates when the user uploads a package
- it gets a (success) message of how many updates were applied when updating content types from the hub
- it gets an (error) message with what exactly is wrong when the validation of an uploaded package fails

The PHP implementation uses a global function to add success and error messages to the AJAX response. I would like to avoid this pattern as it won't work nicely with unit tests. I also think that the domain classes shouldn't know about how they are called, as there are scenarios where they might not be called by an AJAX call at all, but are used to automate things (e.g. in CLI dev tools). Thus, it would be better if these messages were reported back to the top level with another mechanism. Suggestions:

- Option 1: Report messages by sending them as events through a Publish-Subscribe pattern.
- Option 2: Report messages by using a logging solution with log levels (like Winston)
- Option 3: Return success messages as strings and report error messages by throwing error objects.

The top level caller (i.e. the method in the implementation (h5p-demo) dealing with the AJAX request) then adds the success or error message to the AJAX response.

JPSchellenberg:
Correct me if I am wrong, but I think we need to wait for every message before resolving the request and sending the response. So if we use for example Option 1, how do we know that there is no pending message? And if we use Option 3 then we catch just one error and report that error.
How do we check, that there are no pending messages/events before sending the response?

sr258:
> Correct me if I am wrong, but I think we need to wait for every message before resolving the request and sending the response. So if we use for example Option 1, how do we know that there is no pending message? 

In this model, the domain model classes publish events/messages at an aggregator (the aggregator is passed to them on construction). The class dealing with the request subscribes to the messages on the aggregator before it calls the domain model classes. That way every message will be received at the top level and it has to be stored somewhere to collect all messages that the domain model classes emit. Once the domain model's promise has resolved, the calling class adds the messages to the AJAX response. There will be no more pending messages because the promise has resolved.

> And if we use Option 3 then we catch just one error and report that error.
How do we check, that there are no pending messages/events before sending the response?

[...]

Issue #44, opened 2019-08-01
Localization of string literals (of the core)


sr258:
While working on the validation part of h5p-extractor I've come across the need to use string literals that are displayed to the user (the validation errors). These literals should obviously be localized to the user's language.

The PHP library does it this way: Whenever it outputs a message to the user, it calls the function [t($message, $replacements)](https://github.com/h5p/h5p-php-library/blob/4a9ea77d470906ce1c114361493e0c94817c29d9/h5p.classes.php#L62) with the English message string as first parameter. This function returns the translated string and does some placeholder replacements.

This means that the English message string is put in as a literal value into the actual source code. I think this is bad style (changing English literals means you have to look through the whole source code; creating translations is tedious; translations can't take context into consideration, which means that the same English literal in different places always has to have the same translation) and shouldn't be replicated in the NodeJS library. 

What I've done is put the (English) literals into a JSON file as key-value pairs. The JSON file for the user's language is loaded by the framework and the content is passed to the validator as an object. Whenever the validator outputs a literal it looks up the string in the translation object through the key and outputs the value (the translation). If it requires placeholder replacements it does it with a little helper function (see [here](https://github.com/sr258/H5P-Extractor/blob/aae244748db58fa24589ceb0bd2ccf03c31fe7d5/src/validator.js#L145) for an example).

I think my current approach is not the final word on localization of the NodeJS implementation (doesn't cater for single strings that are not translated in a language even though there is a language file for it), but this is something that has to be used project wide, so I would like to check if you've had any thoughts on this.

sr258:
I've gone ahead and introduced a [mockup TranslationService](https://github.com/sr258/H5P-Extractor/blob/validator/test/mockups/translation-service.js) in the validator. It can be injected in the constructor and implemented by every integration. If you like this approach I would extend it to the ContentTypeCache I've already written.

The more I look at it, the less sense it makes to have h5p-extractor as a standalone project. There will always be dependencies that have to be injected by h5p-editor-nodejs-library, so it would make sense to move extraction and validation to the editor.

JPSchellenberg:
I like your idea of a TranslationService and would say we use it. We can also integrate the H5P-Extractor into the Editor as they really seem to be tightly coupled. We definitely should try to avoid mistakes that the PHP-Library does like message strings within the source code.

sr258:
[...]

PR #187, opened 2019-10-17
refactor(types)!: Made naming and interface usage more consistent (closes #119)

### LINKED ISSUE #119: Check Library interfaces
I think the library interfaces right now mix information that is received from the H5P Hub and that is taken from library.json. Sadly, the formats are not exactly the same. The format that is used in library.json is already specified in library-schema.json and should be the one that is used in ILibraryJson. The format we get from the hub is the one in the Library class.

We should also use more telling names like ILibraryJson, HubLibrary and a clearer one for ILibraryData (can't think of one right now).

sr258:
In this PR, the somewhat inconsistent usage of library interfaces and classes has been made consistent:

- Every library interface in types.ts now has a clear and understandable name to differentiate them. There also is a clear inheritance hierarchy between them.
- The library interfaces in types.ts no longer extend the Library class.
- The LibraryManager and FileLibraryStorage classes use the minimal interfaces needed for their functionality. Before, they always used the full Library class, even though they only require a machine name, major version and minor version.

Sorry that the PR changes so much, but there are no smaller steps possible to tackle the issues of inconsistency in this area.

The PR is not fully finished yet. I have to re-check everything, but it's too late for this today.


JPSchellenberg:
That's a great and long overdue PR. I wanted to refactor the types for a long time but postponed it every time. :) 
After we refactored this, how comfortable are you with the API? Do you think we could release it as v1.0.0 after this PR is merged?

sr258:
We're approaching API 1.0, but I don't think we're there yet:
- The storage interfaces don't have all necessary CRUD operations, I think. (e.g. deleting is missing #146 )
- We still have to decide on a consistent way of reporting back errors and success (see #83)
- I think we should also decide where to check permissions (#112). Currently the storage objects have method signatures that imply that the storage provider has to / can check permissions. 
- #44 also has to be settled: Currently all components that emit messages to the user have a dependency on the translation service. It might be better to remove the dependency and simply emit H5pError objects. The actual localization is then done in the presentation layer (H5pEditor). (It's possible that this won't affect the API)

Maybe a phone call is easier on this than to do it all by typing?

sr258:
The PR is ready to be reviewed, I think. I haven't yet split the types.ts file into two to not make the PR even bigger and more confusing.

JPSchellenberg (approved):

JPSchellenberg:
> Maybe a phone call is easier on this than to do it all by typing?


Yeah, I think this is a good idea. I'll contact you in Slack

PR #219, opened 2019-11-01
feat(h5pcore): update h5p-core to v1.24

JPSchellenberg:
Hey,

I started to update the H5P-Core to v1.24, but unfortunately it is not as easy as I hoped. The tests still fail. I will see if I can fix that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,8 +73,8 @@ The H5P client (running in the browser) sends many AJAX requests to the server (
 
 This application doesn't include the H5P JavaScript core files for the editor and the player. These are the files that make up the editor and player that the end user interacts with in the browser. The core files must be obtained seperately:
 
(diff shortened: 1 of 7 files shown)


sr258:
You'll also have to add references to new JS files. See https://github.com/Lumieducation/H5P-Nodejs-library/commit/b449436a33b2d5671996ef4639510610de26f6b1#diff-0f6ad0fcd9170cc1649a8377c9dcee67L76

sr258:
See examples/implementation/EditorConfig#36:
public coreApiVersion: { major: number; minor: number } = {
        major: 1,
        minor: 23
    };

sr258 (approved):
We should do some manual tests whether all functionality of the editor that has worked so far still works, too.

PR #222, opened 2019-11-01
feat(saving): Implemented content scanner

sr258:
I've added a content scanner that goes through the parameters (=content.json) of a piece of content and extracts all references to local files (ContentFileScanner). This is needed for the content saving functionality, to determine which files to copy from temporary storage and to find out which files were removed since the last save. (The saving mechanism can use the JSON paths returned by the scanner to modify the correct objects.)

The ContentFileScanner is based on a more general ContentScanner class, which can be used to get all kind of information from a parameter object.

While the scanner works fine (finds all files in the Hub examples), it is integrated into the file saving process yet. But as this PR would get too big if I added this too, I think it's time to review & merge.


sr258:
Finally, the branch passes all the tests. I've separated the 'integration' tests (= tests using all the hub examples) into a different CI stage with a limited number of worker. We can try to improve memory usage in the future, if this is indeed a problem, but I think optimization should come later.

@JPSchellenberg You can review now.

JPSchellenberg (approved):

Issue #87, opened 2019-09-01
Uploading images or other files broken for unsaved content


sr258:
Currently, uploading images doesn't work if the content you're editing is new and hasn't been saved before. The reason is this would require either having a temporary storage place for unsaved content or the concept of preliminary content in storage (see #61). We have to choose one path and implement it to make uploading images for unsaved content work again.

JPSchellenberg:
I have a really hard time deciding which path to go as I don't know what fits best. I think I would go with the preliminary content. Which path would you prefer?

sr258:
I think having preliminary content is easier to implement (just a flag, an expiration time and a method to call regularily) and it would be easy to transition to the temporary storage approach if we find any reason why preliminary content was a bad idea after all. So I vote for preliminary content. It could be implemented by extending the content of the .h5p file in the content folder we are having already. As far as I understand it this file is not handed out the client anyway, so we can add properties there that indicate that content is preliminary and when it is expired. The .h5p file (with just the preliminary marker and the expiration date) would have to be generated either when the user opens the editor or when he first saves a content file. If we generate it right away this would mean avoiding possible collisions of ids but we might have many preliminary content entries.

JPSchellenberg:
After thinking about this for a while I think we should leave a lot of this problem to the implementation as there might be different scenarios in which one way is better than the other. The [file-content-storage](https://github.com/Lumieducation/H5P-Editor-Nodejs-library/blob/master/src/file-content-storage.js) is relying on the local file-system, but what if someone wants to use amazons S3 storage? I like that the FileManager is interchangeable, but how would it look like exactly?

sr258:
The ContentStorage classes provided by the implementation would be unaware of preliminary content, as this would only be a flag and an date attribute in the metadata of a piece of content. Managing the creation, final storage and deletion of preliminary content would be entirely dealt with in the storage agnostic ContentManager. To the storage creating preliminary content would be the same as a regular piece of content. The actual storage of content when the user saves it would look like a content update. This works long as the storage class can deal with any kind of metadata (as we'd extend what's in h5p.json). 

Switching between the two modes isn't that easy as the different solutions would lead to changes in several places and make the code more complex.

JPSchellenberg:
That sounds good, so I would opt for the preliminary content.

sr258:
[...]

Issue #61, opened 2019-08-15
Use temporary folder for uploaded content files


rtens:
This issue combines #23 and #33.

Currently, when uploading files during editing (eg by adding a picture to a Course Presentation slide), the file is stored in the content folder (under `h5p/content/{contentId}/content/{fileName}`).

This leads to the problem that if the changes are not saved and the file is not used in the end, it will still be part of the package and not be deleted.

A possible solution would be to store uploaded files in a temporary folder and only move them to the content folder when the changes are saved.

The temporary folders should be regularly purged.

sr258:
I've given this some thought and I think it would be better to have preliminary content in the storage instead of having a global temporary storage location from which the files are copied when the corresponding piece of content is finally saved. Every time a user opens the editor, a preliminary content entry is created with a timeout after which it is deleted automatically if it hasn't been saved permanently until then.

- this reduces complexity in the editor module (no need to differentiate btw. creating new content and editing existing content; no need for a special temporary files module that has to keep track of unsaved sessions, manage associations btw. content files, users and sessions etc.)
- this makes saving preliminary content states easier in the future (e.g. regularly saving the state of content.json to recover from crashes/connection drops)

@rtens @JPSchellenberg What do you think of this approach?

rtens:
Not sure if this works (at the the way I understood it) since it's only the content files that are uploaded, the content.json itself is not updated with each file upload. So the preliminary content would be have the new files but the old content.json.

sr258:
I thought of having a database entry without content.json. This should be not problem and you simply add the content.json contents later.

rtens:
My gut feeling leans towards keeping the temporary files separate and copying them over on save. But only slightly.

PR #229, opened 2019-11-02
Feat(saving): Implement saving with temporary content files

sr258:
When this PR is finished, the editor will use the temporary file storage system of #61 and enable uploading files for unsaved content (#87).

Steps:

- [x] reroute all content file saving to temporary storage and return correctly marked file paths
- [x] detect if a requested file is in temporary storage and fetch it from there (requires new method in H5PEditor)
- [x] when saving / updating content: scan parameters of new content and copy required files from temporary storage to permanent storage; change references in parameters
- [x] when updating content: compare old parameters to new and delete files from storage that aren't referenced any more
- [x] add tests
- [x] check edge cases:
   - pressing "save" twice
   - server crashes during saving


sr258:
I think the PR is ready to be merged. I've added documentation calling the cleanup of temporary files regularly, as I've decided to follow the way the PHP implementation handles temporary file deletion when saving content: Temporary files are not deleted when content is saved, but when they are supposed to be cleaned up anyway. The reason for this is that the temporary files should be kept around for cases like when the user has pressed save twice.

The unit tests run fine on my dev machine and take 3.5s, no idea why they take so long in the CI pipeline.

JPSchellenberg (approved):

Issue #92, opened 2019-09-04
Make apiVersion consistent in all parts of the project


sr258:
Currently, the apiVersion reported to other components (hub, client) does not always derive from the same source. 

It can come from the value in default_editor_integration.json or from config.js. I think we should store it in the config and it should be added to the H5PIntegration object dynamically. 

Ideally, the supported coreApi version is determined by deriving it from the version of the core JS files that were downloaded from GitHub and not set manually.

PR #230, opened 2019-11-03
feat: Reduced redudant core api versions and added documentation on core updates

sr258:
I've removed the hardcoded core API in ```default_editor_integration.json``` (see #92) and added documentation on how to update the core files (for future use).


JPSchellenberg (approved):
Very good. :)

PR #247, opened 2019-11-10
fix(example): Fixed reference to default import

sr258:


sr258:
Obsolete as the example is now in TypeScript.

PR #252, opened 2019-11-11
test(examples): convert express-example to typescript

JPSchellenberg:
Hey,

I converted the express example to typescript. @sr258 What do you think?


sr258 (approved):
Everything looks fine. This will simplify development as the IDE will now correctly find references in the example.

One disadvantage of only having a TypeScript example is that JS developers might think that they need TypeScript to use the library, while it's fine to use JavaScript.

Issue #144, opened 2019-10-01
Support for Copy & Paste operations on the server-side


sr258:
Even though Copy & Paste is mostly done in the web client, the server must seem to support cloning image files and the like.

sr258:
Relevant passage in h5p-php-library:
h5p.json:2651ff

PR #254, opened 2019-11-11
Feat(copy-paste): Support copy and paste

### LINKED ISSUE #248: Make use of 'content' path consistent
Content files (images) are currently saved within the 'content' directory of a piece of content. This is not necessary. We should remove this directory and make the use of ContentManager and FileContentStorage consistent that way.

sr258:
When finished this PR will make Copy & Paste work (#144).

- [x] fix content paths (#248)
- [x] deal with relative paths in ContentStorer

Resolves #248, #144

### DIFF
--- a/examples/express.ts
+++ b/examples/express.ts
@@ -64,7 +64,7 @@ const start = async () => {
         stream.pipe(res.type(path.basename(req.params.file)));
     });
 
-    server.get(`${h5pRoute}/content/:id/content/:file(*)`, async (req, res) => {
+    server.get(`${h5pRoute}/content/:id/:file(*)`, async (req, res) => {
         const stream = await h5pEditor.getContentFileStream(
             req.params.id,
             req.params.file,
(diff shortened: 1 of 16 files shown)


sr258:
**This is a breaking change, but we've agreed not to mark those for semver to not reach 1.0 yet.**

Copy & pasting now also works with files: The server now clones files that have relative path in the params (this is how the H5P editor client deals with copy & pasting files).

I've also noticed that the implementation FileContentStorage was vulnerable to an attack with relative paths (you could navigate out of the content folder into the system dir by preparing a content.json file) and fixed this. (We should think about hardening the application in general.)

A change to have ContentManager and FileContentStorage handle paths was necessary (this is the breaking change). Up to now content files were stored in a 'content' directory within the content directory. The URLs for content files also was 
```/h5p/content/contentId/content/file.ext```
The editor so far generated this weird URL, because nodeVersionId was not set to contentId. I've corrected this and now it generates paths like these:
```/h5p/content/contentId/file.ext``` 
The classes mentioned above had to be adapted to match the new URL.

JPSchellenberg (approved):

Issue #257, opened 2019-11-12
Export interfaces


JPSchellenberg:
Hey,

I think the interfaces that have to be implemented by the user should be exported from our module so that you can use 
```
import { IEditorConfig } from 'h5p-nodejs-library';
```
Unfortunately, I could not figure out how to do this properly.
@sr258 any ideas?

jankapunkt:
Would this strictly require typescript for my implementation?

JPSchellenberg:
No. We will always release for plain javascript. Exporting interfaces just enables typescript users to have type checking.

sr258:
https://cameronnokes.com/blog/the-30-second-guide-to-publishing-a-typescript-package-to-npm/ seems like a decent guide. I‘ll do this tomorrow.

PR #280, opened 2019-11-23
feat(exports): Interfaces are now exported with type definitions

sr258:
There now is an index.d.ts file in build/src, which is an entry point into the type definitions needed to write an implementation. I had to remove the default export for this to work, so you must now import the library like this: (breaking change)

```
import * as h5pLib from 'h5p-nodejs-library'
```

The Express example now also doesn't import the source TypeScript files directly anymore, but imports from index.ts. 

Currently, the package also exports a few (utility) classes, which is ok in some cases (H5pError, LibraryName, InstalledLibrary). In others we should try to remove the exports in the future (PackageExporter (should be instantiated internally, see #282), TranslationService (should be implemented by the example)). I don't want this PR to become too big, so I think we can do this later.

Note that there is an "error" in the index.d.ts file, if you install the package:
image
The reason for this is that you need these TS compiler flags for TS to accept JSON files in imports:

```
"resolveJsonModule": true,
"esModuleInterop": true,
```

The error means that the ``englishStrings`` property doesn't have typings, but the values can be read fine, as the actual import is done in the JS file.

### DIFF
--- a/examples/express.ts
+++ b/examples/express.ts
@@ -9,7 +9,7 @@ import util from 'util';
 const exec = util.promisify(child_process.exec);
 import index from './index';
 
-import H5P from '../src';
+import * as h5pLib from '../src';
 
 import DirectoryTemporaryFileStorage from './implementation/DirectoryTemporaryFileStorage';
 import EditorConfig from './implementation/EditorConfig';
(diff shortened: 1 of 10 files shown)


sr258:
Resolves #257

JPSchellenberg (approved):

Issue #278, opened 2019-11-22
Uploaded packages should not create new contentIds


sr258:
**Describe the bug**
When a user uploads a package, we currently create a new piece of content and assign a new contentId right away.

**Expected behavior**
- The libraries in the package should be installed.
- The content files in the package should be added to temporary storage. References in the parameters are changed accordingly.
- The server returns the parameters (=content.json) to the client. (Not sure how this is done.)

PR #289, opened 2019-11-26
fix(package upload): content of uploaded packages is now not permanently stored in the system

sr258:
- [X] change signatures of PackageImporter to return metadata and parameters instead of contentId
- [X] implement logic to copy content files to temporary storage

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,3 +181,4 @@ Read more about them at the following websites:
 -   CARO - https://blogs.uni-bremen.de/caroprojekt/
 -   University of Bremen - https://www.uni-bremen.de/en.html
 -   BMBF - https://www.bmbf.de/en/index.html
+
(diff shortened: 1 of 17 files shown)


sr258:
Turns out the POST request to the 'filter' Ajax endpoint invokes some method of H5PContentValidator in the PHP implementation, which is a 1340 lines long class doing lots of stuff...

I don't want to implement this right away, but have a closer look at what this is doing. At the moment, I simply don't "filter" the parameters but return them. Uploading packages works fine, though, but maybe things aren't as safe as we want them to be.

sr258:
Closes #278

sr258:
Rerunning the tests worked... weird.

JPSchellenberg (approved):

PR #298, opened 2019-12-02
feat(implementation): add default fs-implementation (closes #284)

### LINKED ISSUE #281: Cleaning up example
I think the Express example has become quite big - I'd like to refactor it and also remove parts (e.g. copying examples from the h5p.org site - we have a working h5p hub after all). @JPSchellenberg would you agree?

### LINKED ISSUE #284: Add default file implementation
**Is your feature request related to a problem? Please describe.**
I would like to have certain 'default implementations' like the file-implementation used in our example. 

**Describe the solution you'd like**
I would want to be able to import a default file-implementation like
```
import * as H5P from 'h5p-nodejs-library';
const h5p = H5P.fs(
    editorConfig: IEditorConfig,
    librariesPath: string,
    contentPath: string,
    temporaryStoragePath: string
);
```

**Describe alternatives you've considered**
We could add this to this to this library or make another repo/npm-package.

@sr258 what do you think?

JPSchellenberg:
Hey,

as discussed in #284 I added a implementation which uses the file-system as storage. The implementation is just a wrapper around the implementation used in examples/ that @sr258 has written.
I know that you were against a implementation of the contentStorage, which is why I added the option to provide a custom contentStorage instead of a path.
My main goal with this is to get an easy and fast `getting started`, as I think in the current state it is quite difficult to get started with our library. Another way could be to export the different storage classes and do something like:
```ts 
const h5p = H5P.fs(editorConfig, H5P.FileLibraryStorage('/path/to/libraries'), H5P.TemporaryStorage('/path/to/tmp'), H5P.FileContentStorage('path/to/content'));
``` 

instead of the [current interface](https://github.com/Lumieducation/H5P-Nodejs-library/blob/feat/fs-implementation/examples/express.ts#L21).

What do you think?


src/implementation/fs/InMemoryStorage.ts
```diff
@@ -1,10 +1,8 @@
-import { IKeyValueStorage } from '../../src';
-
 /**
  * Stores objects in memory. It can store any key-value pairs.
  * This is just a placeholder for a proper storage implementation.
  */
-export default class InMemoryStorage implements IKeyValueStorage {
+export default class InMemoryStorage {
```
sr258: Why remove the interface? I think we still need it as the in memory storage is mostly there to cache the h5p hub info. In other scenarios, there might be a Redis cache or something that implements IKeyValueStorage.

src/implementation/fs/index.ts
```diff
@@ -0,0 +1,27 @@
...
+    contentStorage?: IContentStorage
+): H5P.H5PEditor {
+    return new H5P.H5PEditor(
+        new InMemoryStorage(),
+        editorConfig,
+        new FileLibraryStorage(librariesPath),
+        contentStorage || new FileContentStorage(contentPath),
+        new H5P.TranslationService(H5P.englishStrings),
+        (library, file) =>
+            `${librariesPath}/${library.machineName}-${library.majorVersion}.${library.minorVersion}/${file}`,
```
sr258: The resolver here must return a URL at which the library can be retrieved. This will return the local path. I don't think the resolver can be part of the fs implementation.

src/implementation/fs/index.ts
```diff
@@ -0,0 +1,27 @@
...
+    temporaryStoragePath: string,
+    contentPath: string,
+    contentStorage?: IContentStorage
+): H5P.H5PEditor {
+    return new H5P.H5PEditor(
+        new InMemoryStorage(),
+        editorConfig,
+        new FileLibraryStorage(librariesPath),
+        contentStorage || new FileContentStorage(contentPath),
+        new H5P.TranslationService(H5P.englishStrings),
```
sr258: I think the TranslationService should be constructed in H5PEditor using a default if no service is passed to the constructor. The fs implementation is not really the best place for it.

examples/express.ts
```diff
@@ -11,28 +11,18 @@ import index from './index';
...
-            new JsonStorage(path.resolve('examples/config.json'))
-        ).load(),
-        new FileLibraryStorage(path.resolve('h5p/libraries')),
-        new FileContentStorage(path.resolve('h5p/content')),
-        new h5pLib.TranslationService(h5pLib.englishStrings),
-        (library, file) =>
-            `${h5pRoute}/libraries/${library.machineName}-${library.majorVersion}.${library.minorVersion}/${file}`,
-        new DirectoryTemporaryFileStorage(path.resolve('h5p/temporary-storage'))
+    const h5pEditor = h5pLib.fs(
+        new EditorConfig(),
```
sr258: We need persistence of the config because the site must store its uuid somewhere. Otherwise we don't conform to the H5P Hub protocol and retrieve a new uuid everytime the server restarts.

src/implementation/fs/index.ts
sr258: You can implement it in the express middleware you're writing though.

sr258 (changes requested):
I've left a few comments in the code. I indeed think the implementation for developers is simplified that way, but we strongly push towards using the file implementations. They're certainly not fully production ready (in particular FileContentStorage) and we should check them thoroughly.

How about we name the implementation simpleFs, localFs or something like this to make clear that the implementation is not complete?

I think the docs must be updated, too.

sr258 (dismissed):

sr258:
I've merged in #299 and added all the changes I requested above. @JPSchellenberg you can review now.

sr258 (approved):
You've done a great job @sr258 ! ;)

PR #299, opened 2019-12-02
feat(adapter): add express-adapter (closes #281)

### LINKED ISSUE #281: Cleaning up example

JPSchellenberg:
Hey,

I started an adapter for express. It should make the implementation easy and look something like this: 
```ts
    server.use('/h5p', H5P.adapters.express(h5pEditor, path.resolve('h5p')));
```
I still have to work on this and while I am on it, I am making the example clearer and leaner.


sr258:
I took the liberty to continue working on this, as I'd like to have the Express adapter for a test tool using the library. Now the editor and player seem to work all right, but I think we should think about the name of the routes used by the adapter and how URLs referencing these routes are generated:

- The routes are not consistently structured (sometimes the contentId is part of the URL sometimes it's a query parameter) and named.
- _(The paths to routes (e.g. '/libraries', '/content', '/download') should derive from a single-source-of-truth and be generated in the editor. This is something we should look into in general (I've removed the unneeded URL resolver for library files, as we hard-code that stuff anyway, but we should derive all routes from the config).)_ (done, see a1e9f68)

sr258:
Now the URLs of the express routes are all generated dynamically from the config and I've made the endpoints more consistent:

I'm happy with these paths:
```
/h5p/libraries/H5P.Example-1.0/abc.js
/h5p/content/1234/image.png
/h5p/temp-files/image.png
/h5p/params/1234
/h5p/ajax?...
/h5p/core/...
/h5p/editor/...
/h5p/download/1234
```
Those endpoints will not vary from implementation to implementation, so it makes sense to integrate them like this.

The endpoints that display pages, however, should be moved to a different place in my opinion:
```
/h5p/play/1234
/h5p/edit/1235
/h5p/new
```
What these endpoints return might vary from implementation to implementation. Maybe we should move them into a different router and put them into the example? Of course, the renderers already take care of customizing the output, but I'm not sure if we should also move the endpoints somewhere else. After all the implementation, might not want to use the `/h5p/...` URL to display pages. @JPSchellenberg what do you think about this?

I haven't thought of a new new main example page, but I'll do so tomorrow.

JPSchellenberg:
Hey, sorry, I was a little bit off the last weeks. I completely agree with you. These routes 
```
/h5p/play/1234
/h5p/edit/1235
/h5p/new
``` 
depend on the implementation. We should focus on the ones that do not vary. I will have a look into this PR today. Do you consider it still as work in progress?

sr258:
No worries. Had lots of things to do at work, too. 

The PR is mostly finished. I didn't come round to adding proper tests, yet. The overall coverage has decreased in the last few months (mostly because I moved non-code objects into JSON files which don't count towards the coverage), so I thought it might be good to add more tests here.

However, I think it can be reviewed and also merged.

sr258:
The express adapter is also not yet separated into the generic and non-generic parts (play/edit/new routes).

sr258:
I didn't add new tests, but I've moved all the non-generic views into their own Express adapter. I think the branch can be reviewed and merged now, @JPSchellenberg .

sr258:
I've merged this branch into #298 and added the changes I requested in December. You can disregard this PR and merge #298 instead.

PR #300, opened 2019-12-03
docs: Added GitHub Action to generate coverage report and send it to Coveralls

sr258:
I've added an GitHub action that automatically runs when we push to the repo or create a PR. It generates a coverage report and sends it to Coveralls (free for OpenSource). That way we get test coverage badges and reports (I've tested this on my fork and it works fine):
The badge we would get:
image

A CI pipeline report on coverage changes due to PRs/pushes:
image

The details report on coveralls.io (here you can browse through the source and look at the uncovered lines):
image

P.S. I think we should consider switching to GitHub Action alltogether: it has up to 7 GB of RAM (we have had memory issues...) and we can test on Linux, Windows and Mac!

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,27 @@
+on: ["push", "pull_request"]
+
+name: Coveralls coverage generator
+
+jobs:
+
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+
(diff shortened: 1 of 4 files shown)


JPSchellenberg (approved):

JPSchellenberg:
That sounds great! Switching to GitHub Actions sounds good. I like to have everything in one place.

PR #313, opened 2019-12-05
fix(ContentTypeCache): localContentType is converted to HubContentType (fixes #308)

### LINKED ISSUE #308: TypeError: localContentType[0].canBeInstalledBy is not a function
```
TypeError: localContentType[0].canBeInstalledBy is not a function
  File "/srv/node_modules/h5p-nodejs-library/build/src/ContentTypeInformationRepository.js", line 159, in ContentTypeInformationRepository.<anonymous>
    if (!localContentType[0].canBeInstalledBy(user)) {
  File "/srv/node_modules/h5p-nodejs-library/build/src/ContentTypeInformationRepository.js", line 33, in step
    op = body.call(thisArg, _);
  File "/srv/node_modules/h5p-nodejs-library/build/src/ContentTypeInformationRepository.js", line 14, in Object.next
    function verb(n) { return function (v) { return step([n, v]); }; }
  File "/srv/node_modules/h5p-nodejs-library/build/src/ContentTypeInformationRepository.js", line 5, in fulfilled
    function fulfilled(value) { try { step(generator.next(value)); } catch (e) { reject(e); } }
  File "internal/process/next_tick.js", line 68, in process._tickCallback
```

JPSchellenberg:
Hey,

as discussed in #308 I fixed the problem, that a localContentType was not properly converted to a HubContentType and therefore was lacking the `.canBeInstalledBy`-method. This happened if a ContentTypeCache was loaded from storage.


src/ContentTypeCache.ts
```diff
@@ -172,11 +172,15 @@ export default class ContentTypeCache {
         if (!machineNames || machineNames.length === 0) {
             return cache;
```
sr258, line 173: Won't we need to make sure that the function is attached if all contentTypes are requested? (The return value you've changed below is only used in certain cases if you look for a specific contentType).

sr258 (changes requested):

sr258:
A test for the error might also be good.

src/ContentTypeCache.ts
JPSchellenberg: You are completely right. My bad. I will change this today.

JPSchellenberg:
> A test for the error might also be good.

I will try to write one today.

JPSchellenberg:
Ok, I added a test and made sure that all returned HubContentTypes are valid.

sr258 (dismissed):
The test didn't work out the way it was supposed to, as the cache is not part of the EditorConfig but stored aside it. It also doesn't store the raw format returned from the HTTP endpoint but a transformed version. I've changed the test accordingly.

JPSchellenberg:
But now the test also passes even if HubContentType conversion is removed. This caused due to the fact, that the ContentTypes are [converted](https://github.com/Lumieducation/H5P-Nodejs-library/blob/master/src/ContentTypeCache.ts#L141) to the correct HubContentType when downloaded via `.forceUpdate()` and stored in memory. The memory saves it as HubContentType-object with all member-methods.

The problem only occurs if the contentTypeCache is loaded directly from a .json or other storage as a .json and a database will store it as an object without the member-methods - therefore `.canBeInstalledBy` is not defined.

Image an Editor without access to the ContentTypeHub. It loads all its ConentTypeCache from storage other than memory. And in this case, we have to convert the data loaded from storage to the "internal-format" as well.

The test should not access the ContentTypeHub and load its data only directly from a persistent storage (in this case JsonStorage).

sr258:
Ok, I see. The problem is that we still have to store the internal format of the cache and not the one we receive from the public http endpoint. 

This means, we have to update the cache regularly, persist the cached data from IMemoryStorage to JSON (JSON.stringify), convert it back into an object, put this into a new InMemoryStorage, construct a new ContentTypeCache and then call the get() method.

JPSchellenberg:
But that sounds complicated. As far as I see the only difference between the ContentTypes from the Endpoint and the ones stored in the storage is the missing `.canBeInstalledBy`-method. So by converting the ContentTypes loaded to the internal-format everything works fine: https://github.com/Lumieducation/H5P-Nodejs-library/pull/313/commits/a6c2ce9a06688563f4dfb77fd392ebb55820b7f7

```ts
let cache = (await this.storage.load('contentTypeCache')).map(
             ContentTypeCache.convertCacheEntryToLocalFormat
         );
```

JPSchellenberg:
Why does the ContentTypeCache have to be in an InMemoryStorage? Am I missing something?

sr258:
The internal format is somewhat different from the raw one and the conversion is done here (it's not just the missing method):
image

You don't have to use InMemoryStorage, but that way we avoid writing a temporary file to the disk in the test. In a real-life implementation I would want to keep the cache in memory and only fall back to the file as a fallback. (This would be yet another implementation of the IKeyValueStorage interface.)

As a solution we could simply make canBeInstalledBy() static and make HubContentType an interface. That way we don't have to worry where the object is from and whether it has a method or not.

JPSchellenberg:
Ok, I see. I like your solution. But I would not keep the cache in memory as it gets difficult when you want to use multiple instances of the library. Or are you also referring to something like redis as memory?

sr258:
In a single instance a simple the InMemoryStorage is fine (this will be 90% of the installations), in a cluster environment we'd need redis, yes.

JPSchellenberg:
I converted the HubContentType into an interface and now it uses the already exists .canInstallLibrary-method of the ContentTypeInformationRepository, which seems to have the same logic. And it is already tested [here](https://github.com/Lumieducation/H5P-Nodejs-library/blob/master/test/ContentTypeInformationRepository.test.ts#L247).

JPSchellenberg:
Now the tests pass. So it was my code. I am sorry, I did not realize that almost all tests are already mocked. Great work.

sr258 (approved):

PR #321, opened 2019-12-13
fix(findlibraries): prevent TypeError object[key].match is not a function (fixes #320)

### LINKED ISSUE #320: TypeError: object[key].match is not a function
**Describe the bug**
The H5P-Nodejs-library throws the error `TypeError: object[key].match is not a function`

**Environment (please complete the following information):**
 - OS: all
 - Browser all
 - Nodejs-Version 10.17.0

**To Reproduce**
1. Go to http://test.Lumi.education/
2. Click on `Open Existing H5P`
3. Click on `h5p` in File Browser
4. Click on `examples`
5. Click on `H5P.Questionaire.h5p`

**Expected behavior**
H5P.Questionaire.h5p should be imported & this error should not be thrown

**Screenshots**
Screenshot 2019-12-13 at 12 51 18

**Additional context**
H5P.Questionaire.h5p is imported correctly. The problem occurs when `.saveH5P()` is called with the parameters.

JPSchellenberg:
Hey,

as mentioned in #320 some content-types can not be saved via `H5PEditor.saveH5P()` as the libraries can not be determined using the .`findLibraries()`-method.

I wrote a test to test which content-types are affected.


JPSchellenberg:
Ok, I have no idea why the coverage-test fails.
The Validator fails because it can not find the `semantics.json` of some libraries. I guess they get deleted when I use the hub-content-folder as [library-folder](https://github.com/Lumieducation/H5P-Nodejs-library/pull/321/files#diff-a9846efc20e3e2e167b060b96af78e0bR39) for the FileLibraryStorage in my test. But what I don;t understand is why. I copied the extracted content to a temporary folder and use it there. The other tests should not be affected. @sr258 any ideas?

sr258:
The error is really weird, as the stack trace shows that the exceptions are thrown in the H5PEditor.saveH5P() test. Jest however reports test failures at other test modules...

Normally these 'no such file or directory' errors occur if you don't await a promise somewhere and the files are cleaned up too early by withDir. This could be caused by  the call to `contentTypes.forEach`. You have async functions in there and never await them. So withDir cleans up the directories too early. `await Promise.all(...)` is the better choice here. Alternatively, you can also iterate over it('...). This makes it easier to spot which file failed. I've done this in the integration tests. Check out PackageExporter.test.ts for an example.

Btw: Why manually copy all the files from the packages into a special directory? The package upload functionality works fine after all, so you could simply call h5pEditor.uploadPackage(...) This will NOT save the package anymore and just upload all the data files into temporary storage. It returns the metadata and parameters that you can use to save the file. That way we also don't have to extract all the files when running a test.

Could you move the test into the integration folder and only run it as an integration test? If we run the test for every file, this will increase test time a lot and cripple the IDE again.

JPSchellenberg:
Thank you. I made your suggested changes and it seems to work. I also pushed my last commit to the branch, but this PR does not pick it up... I will wait a little bit and see if this PR updates. If it doesn't update I will make a new one tomorrow.

JPSchellenberg:
Thank you. I made your suggested changes and it seems to work. I also pushed my last commit to the branch, but this PR does not pick it up... I will wait a little bit and see if this PR updates. If it doesn't update I will make a new one tomorrow.

translate-h5p:
I think extract-hub-content.ts is not needed anymore, so I've removed it. I've also replaced the redundant manual parsing of the ubername with the function in LibraryName.

sr258 (approved):
I think extract-hub-content.ts is not needed anymore, so I've removed it. I've also replaced the redundant manual parsing of the ubername with the function in LibraryName.

I think this is fine now.
Sidenote: coverall is really pedantic about code coverage decrease. I have introduced a threshold of 0.05%.

PR #363, opened 2020-02-17
Feat(ajax responses): Improved localization handling, error reporting and success messages

### LINKED ISSUE #44: Localization of string literals (of the core)

### LINKED ISSUE #80: Make all error messages localizable
All errors in the module should to be localizable, so we have to move from just throwing text errors to using localizable ids.

I suggest throwing a custom error object (e.g. H5pError) on critical errors that contains a (human-readable) error code an associative array with relevant variables. (see #44)

### LINKED ISSUE #83: Reporting back status with proper AJAX responses

sr258:
The PR closes #44, closes #80 and closes #83. To standardize localization it introduces a dev dependency on i18next (very common npm package) and converts the custom replacement variables (%xyz, @xyz) used so far into more widespread {{xyz}} notation.

Implementations can use any localization library, but must add the property t(...) to the Express req object of the if using the Express adapter. They must also follow the {{...}} notation.


sr258:
The changes turned out to be huge, but now error handling is much improved and the ajax endpoint is more stable and nearly fully tested. This means we now have integration tests of nearly all AJAX endpoints (I didn't test installing libraries from the H5P Hub as we still can't mock downloads from there).

I had to change the interface of the library (breaking changes):
- Removed TranslationService from constructor of H5PEditor (the translation service has been removed)
- ILibraryStorage.getFileStream is now async (needed to check if files actually exist; if you don't do this, the library will throw the file system errors and stop execution in some cases)

The example now also adds i18next to the request object. I think it's good to stay localization library agnostic and leave it to the implementation what library to use. Do you agree, @JPSchellenberg ?

@JPSchellenberg you can review now.

sr258:
I've merged the master

PR #388, opened 2020-03-12
Consistency improvement

sr258:
I'm currently checking the consistency and design of the exported interfaces that implementations must implement. All the changes will be breaking, obviously. I hope they will be the last ones before we have stable interfaces for a longer time, however.

So far I've improved ILibraryStorage (breaking changes).

The following questions have arisen so far:

- Should we use "list" or "get" for methods that return lists of items? Examples in ILibraryStorage: getInstalledLibraryNames, getLanguages, listFiles. Examples in IContentStorage: getContentFiles, listContent. I'd opt for 'list' if we return a list that only includes names or references (e.g. filenames or ids). 'Get' should only be used when we retrieve "full" complex objects.
- Should we use "load" or "get" for object retrieval? Examples: loadLanguage, loadLibrary, loadSemantics, getContentFiles. I'd opt for using 'get' if it is a generic retrieval of data. Load should only be used for direct access to files or databases.
- ILibraryStorage and IContentStorage have dual uses: They deal with libraries and files in libraries, as well as content objects and files in content objects. It would be an option to extract ILibrary and IContent from the interfaces and put all methods that deal with files INSIDE a library or a piece of content in them. This would be more elegant and make interfaces more lean. However, it might also make implementation harder and possibly have negative effects on performance (You'd have to retrieve the IContent object from the storage before you can call methods on it. The extra retrieval of the object might be an extra step if we check for the item's existence. On the other hand we already check the item's existence in the methods themselves, which would become obsolete.)

@JPSchellenberg what do you think?

I'll add more thoughts here, when I look over the rest of the public-facing interfaces.


sr258:
To ease the review, this is a UML diagram of all current interfaces and classes:

5eb9f63d

sr258:
I've now also aligned the H5PPlayer public methods with the H5PEditor style to be consistent. IContentStorage is now also fully consistent with ILibraryStorage (everything that can be added to the storage can also be retreived by it; all methods are now async). Again, all breaking changes.

@JPSchellenberg If you're interested in reviewing the new interfaces, it'd be great if you did this soon. I'd like to merge on Monday or Tuesday, as someone wants to write an implementation against the new interfaces.

PR #503, opened 2020-05-12
Type script usability

BoBiene:


src/H5PEditor.ts
```diff
@@ -60,12 +62,12 @@ export default class H5PEditor {
         public config: IH5PConfig,
         public libraryStorage: ILibraryStorage,
         public contentStorage: IContentStorage,
-        public temporaryStorage: ITemporaryFileStorage
+        public temporaryStorage: ITemporaryFileStorage,
+        private urlGenerator: IUrlGenerator = new UrlGenerator(config)
```
sr258, line 89: Could you please add this new parameter to docs/h5p-editor-constructor.md? I think it would be great if you explained what a typical use case for using a custom URL generator is.

src/types.ts
```diff
@@ -1,5 +1,6 @@
 import { ReadStream } from 'fs';
 import { Stream, Readable } from 'stream';
+import UrlGenerator from './UrlGenerator';
```
sr258: This import seems to be unused.

sr258 (changes requested):
Thank you for your PR! 

I have two small change requests (adding the new parameter of H5PEditor to the documentation and removing an unused import) and would like to understand your use case better: I see that you want to improve subclassing the file content storage classes and this is fine with me. Can you explain your use case in which the protected getter is necessary instead of simply accessing the field directly? (Same with the base url in the UrlGenerator.)

sr258:
@BoBiene I'd be willing to merge this, if you address the two issues I've mentioned and explained why you need the protected getters.

BoBiene:
@sr258 Sure, thank you. 
I'm working on them. 
Currently I have a working prototype of my implementation and as soon as it is all on place I will update this Pullrequest.

src/adapters/expressErrorHandler.ts
```diff
@@ -16,24 +18,16 @@ import H5pError from '../helpers/H5pError';
...
+        res: Response,
+        next: NextFunction
     ): Promise<void> => {
         let statusCode = 500;
         let statusText = '';
         let detailsList;
         let clientErrorId = '';
 
         if (err instanceof H5pError) {
-            if (
```
sr258: Was this removed deliberately? I think we still need it to switch the language to the current user's settings. I have already wondered whether it might not be better to have the implementation call changeLanguage and remove the language override. Any thoughts on this?

If we remove the languageOverride, the currying function in l. 16 must be removed as well, as it doesn't make any sense otherwise.

src/adapters/expressController.ts
```diff
@@ -92,34 +107,32 @@ export default class ExpressH5PController {
...
-            LibraryName.fromUberName(req.params.uberName),
-            req.params.file
+        const lib = LibraryName.fromUberName(req.params.uberName);
+        const [stats, stream] = await Promise.all([
+            this.h5pEditor.libraryManager.getFileStats(lib, req.params.file),
+            this.h5pEditor.getLibraryFileStream(lib, req.params.file)
+        ]);
+
+        this.pipeStreamToResponse(
+            path.basename(req.params.file),
```
sr258: I think you need to pass in the mimetype of the file, but you pass in the filename. I receive warnings in the browser's dev console because of this and some stylesheets are not rendered correctly. 

We already use the npm package mime-types, so you can use it to determine the mimetypes.

src/adapters/expressController.ts
```diff
@@ -92,34 +107,32 @@ export default class ExpressH5PController {
...
             req.user
         );
-        stream.on('end', () => {
-            res.end();
-        });
-        stream.on('error', (err) => {
-            res.status(404).end();
-        });
-        stream.pipe(res.type(path.basename(req.params.file)));
+        this.pipeStreamToResponse(path.basename(req.params.file), stream, res);
```
sr258: See above.

sr258 (changes requested):
Thank you for you extensive PR! I think it indeed improved the developer friendliness of the project a lot, especially for newcomers! 

I've taken the liberty to directly correct a few documentation issues and push the changes. The problem with the mimetypes must be resolved before it can be merged and you should double-check whether you really wanted to remove the language switch in the Express error handler.

src/adapters/expressErrorHandler.ts
BoBiene: It was a mistake while fixing the merge-conflict.
I will correct this.

src/adapters/expressController.ts
BoBiene: i see. fixed that.

sr258 (approved):
I think everything is in order now. Thank you again!

BoBiene:
@JPSchellenberg & @sr258 thank you.

PR #513, opened 2020-05-14
feat(localization): added localization of editor

sr258:
Once the PR is finished, the editor can be fully localized (as far as H5P permits it).


sr258:
I'd consider the PR finished from my point of view. The editor is now fully localizable and I've also added the possibility to customize the metadata form that is displayed for content and resources (at least you can change the licenses). @JPSchellenberg if you want to review it, you can do so now. I'm going to wait a bit and review it myself if you don't have the time.

PR #544, opened 2020-05-22
ci(github actions): added GitHub Action for ci tests

sr258:


sr258:
The PR removes the old CircleCI pipeline and replaces it with a GitHub Actions pipeline. We don't have to mix the two CI systems (coverage generator already uses GitHub Actions) that way. It's also free to test the code on Windows and macOS with GitHub Actions. 

There are some changes how things work:

- To add caching for the h5p packages downloaded from the H5P Hub, it was necessary to remove the auto-download of test files. It is now an extra step that must be called manually. The H5P has a spotty reputation of not being available in high load times. By caching the assets we avoid failed pipelines. @JPSchellenberg Do you think it's worth introducing the extra step?
- The tests are now not run in parallel anymore, but sequentially. This reduces cluttering, as we now have three runners for the different OSes.

While playing around with GitHub Actions when trying to get this working, I've noticed that the GitHub runners seem to be noticably slower than the ones offered by CircleCI. This mostly seems to be the case for the Windows and MacOS runners, which we don't even have on CircleCI right now.

sr258:
I'm giving up on this PR. GitHub Actions is very slow on Windows and there is a weird error with teh e2e for H5P.Audio, which I can't fix. Let's stick with CircleCI.

PR #558, opened 2020-05-27
Dockerfile created and readme updated

jdavidavendano:
Now there is an easy way to run the demo on a docker container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+FROM node:alpine
+
+RUN mkdir -p /home/node/app/node_modules && chown -R node:node /home/node/app
+
+RUN mkdir -p /usr/local/lib/node_modules && chown -R node:node /usr/local/lib/node_modules
+
+WORKDIR /home/node/app
+
+COPY . .
+
+RUN npm install -g typescript
+
+RUN npm install
+
+RUN npm build
+
+USER node
+
+EXPOSE 8080
+
+CMD [ "npm", "start" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,21 +1,15 @@
-FROM node:alpine
+FROM node:lts
 
-RUN mkdir -p /home/node/app/node_modules && chown -R node:node /home/node/app
+RUN mkdir -p /opt/h5p
+WORKDIR /opt/h5p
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+WORKDIR /home/node/app
+
+COPY . .
+
+RUN npm install -g typescript
+
+RUN npm install
+
+RUN npm build
```
sr258: I think the build command is redundant, as `npm install` already performs a build.

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+RUN mkdir -p /home/node/app/node_modules && chown -R node:node /home/node/app
+
+RUN mkdir -p /usr/local/lib/node_modules && chown -R node:node /usr/local/lib/node_modules
+
+WORKDIR /home/node/app
+
+COPY . .
+
+RUN npm install -g typescript
```
sr258: Typescript should be installed locally with `npm install`. Is it really necessary to install it globally, too? I think all my dev machines have typescript installed globally, so I don't know if it is really needed. Could you confirm? If it is not needed, l. 5 would also have to be removed.

sr258 (changes requested):
Thank you for your PR. I think a Docker setup can be useful to try out the demo. We might also consider uploading the demo to the Docker Hub automatically on release, @JPSchellenberg .

Could you please check the two lines I've marked in the Dockerfile? I haven't run the docker build myself, so could you verify whether you really need those lines?

Dockerfile
jdavidavendano: You're right, thanks for the comments, I made the changes.

jdavidavendano:
You mentioned that `npm install` performs a `build` but I've had some trouble with the downloads: 
```
npm run download:content-type-cache
npm run download:content
npm run download:core
```
Those failed in `npm install`
So... I made them again in the Dockerfile and after that, I did a `npm run build`. 
You can download and test the docker container in [https://hub.docker.com/r/jdavidavendano/lumi-education-h5p-nodejs-library](https://hub.docker.com/r/jdavidavendano/lumi-education-h5p-nodejs-library)

sr258:
Thank you for your PR. I've changed the file to use a non-root user. That way the download and build scripts are run on npm install.

PR #833, opened 2020-10-09
fix(content type support): fix for Guess the Answer

### LINKED ISSUE #754: error in Guess The Answer content type
In the player, you get this console error: Cannot read propert 'on' of undefined

Example: https://apps.zum.de/sites/default/files/h5p/exports/interactive-content-2260.h5p

It is caused by an empty params field in an otherwise empty media field. The PHP implementation seems to filter this out, so it is not present in the filtered params at all.

### LINKED ISSUE #755: Error in Column content type
In the player, you get this console error: TypeError: Cannot read property 'split' of undefined

Example: https://apps.zum.de/sites/default/files/h5p/exports/interactive-content-2367.h5p

Cause:
 "content": { "params": {} } 

split operates on the library property of content

sr258:
Closes #754 . Closes #755 .

The error was caused by an empty `media` entry that is not allowed in the player of Guess the Answer, but produced by the editor. The server now has **very basic** library validation functionality that filters out these cases.


src/ContentScanner.ts
```diff
@@ -77,46 +98,62 @@ export class ContentScanner {
...
         switch (elementSemantics.type) {
             case 'library':
-                // If an element contains another library, we have to retrieve the exact name,
-                // and the nested content parameters.
-                if (elementParams.library === undefined) {
+                // If an element contains another library, we have to retrieve
+                // the exact name, and the nested content parameters.
+                if (
+                    elementParams.library === undefined ||
+                    !elementParams.params
```
sr258: Should be removed.

src/types.ts
```diff
@@ -921,6 +921,10 @@ export interface ISemanticsEntry {
     /**
      * (in lists only) defines a single field type in the list
      */
+    extraAttributes?: string[];
```
sr258, line 924: Wrong comment order.

sr258:
Code has to be cleaned up before it can be merged.
I don't like the fact that SemanticsEnforcer manipulates the structure of the params as a side effect. The params might be made immutable and we can return a new tree, or we can make this transparent by naming conventions.

sr258:
Also closes #755

PR #935, opened 2020-11-29
feat(html export): added html exporter

sr258:
The export is not finished yet (but it works) and it might be better to put it into its own package, as there are some heavy dependencies in it (postcss and uglify-js).


sr258:
The export itself works now and the automatic tests pass for all hub content types. I haven't tested all content types manually yet, however.

I've put the HtmlExporter specific dependencies (postcss, postcss-clean, postcss-url, postcss-safe-parser, uglify-js) in the dev dependencies to avoid having them in projects that don't need them. I'm not entirely sure if the package will transpile in typescript if these dependencies are not installed, though. @JPSchellenberg Any guess about this?

JPSchellenberg:
The package does not compile when these dependencies are not installed. Further more I think it will not work during runtime, when these dependencies are not installed. I would suggest to put them into the depedencies and fix that issue later when we are divinding the package into multiple packages.

sr258:
@JPSchellenberg fixed that. It should be ready to be merged.

JPSchellenberg (approved):
This is a really great feature!

PR #1045, opened 2021-02-10
Add xAPI callbacks to h5p-webcomponents & h5p-react

JPSchellenberg:
Hey,

I tried to add the xAPI callbacks to h5p-webcomponents & h5p-react, but I get the error that there is no custom event called 'xAPI' on the h5p-webcomponent.
Do you have any hints?

Screenshot 2021-02-11 at 00 55 49


sr258:
I've fixed this, was just a missing wrapper. 

The current implementation of the xAPI listener will fire for all H5P content objects in the window. As there can be several, an xAPI event in a single content object would be reported by all components. I think the only way around here is to filter the xAPI statement by the contentId of the object. You can use h5p-rest-example for a use case in which there are several player on the same page.

sr258:
@JPSchellenberg What's the status current here? Can I help out somewhere?

sr258:
Strange that there still are conflicts: This PR https://github.com/Lumieducation/H5P-Nodejs-library/pull/1074 doesn't have this issue. I think we should use this one.

PR #1060, opened 2021-02-17
docs(types): added types for user state endpoints

sr258:
Also removed extremely time-intensive test from push


sr258:
@JPSchellenberg everything should work now

JPSchellenberg (approved):

PR #1063, opened 2021-02-17
fix(h5p-server): missing translations when language was not found

JPSchellenberg:


sr258:
Everything's fine. Maybe you point the Turkish users to https://translate-h5p.tk/ so that they can improve the missing translation there.

sr258 (approved):

PR #1067, opened 2021-02-18
Pull back release info from release branch

JPSchellenberg:


sr258 (approved):
We should remove the protection of master to automate this, I think. We're admins anyway, so the protection doesn't make much sense.

PR #1074, opened 2021-02-19
Webcomponents/xapi callback2

sr258:


JPSchellenberg:
I moved the h5p-rest-example docs and made the callbacks fail-safe. Do you think it can be merged now? @sr258

PR #1076, opened 2021-02-20
docs(gitbook): restructuring

sr258:


sr258:
@JPSchellenberg I've removed all the broken formatting caused by GitBook and made the existing documentation more accessible. GitBook works great with it: https://docs.lumi.education/v/docs%2Fstructure/

We should really **not** edit the documentation in the GitBook editor, it seriously wreaks havoc with the markdown files and makes them super ugly.

JPSchellenberg:
This looks really great!

PR #1082, opened 2021-02-22
feat(h5p-server)!: CSRF tokens in UrlGenerator

sr258:
The UrlGenerator can now generate CSRF tokens for the current user.
The editor and player correctly set user objects in the integration (needed for xAPI to work).

BREAKING CHANGES:
The signature of H5PPlayer.render(...) was changed. You now need to pass in a user object and the optional parameters were moved to a dedicated options object.

### DIFF
--- a/docs/advanced/localization.md
+++ b/docs/advanced/localization.md
@@ -35,8 +35,8 @@ shows where this must be done:
 
 | Place | What to do |
 | :--- | :--- |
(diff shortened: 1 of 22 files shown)


sr258:
@JPSchellenberg We need this functionality to have complete CSRF protection of POST routes in lumi-teams. The user object is also needed to generate correct xAPI statements in the player.

JPSchellenberg (approved):

PR #1110, opened 2021-02-26
Update the player events to include  xAPI event

0xMurage:
Updates the doc to include the xAPI event from the player added on `v7.1.0`


sr258:
Thanks a lot for your contribution!

PR #1118, opened 2021-02-28
feat(h5p-mongos3): added MongoLibraryStorage

sr258:


sr258:
Needed to work around the Scaleway rate limiting on S3 requests.

PR #1148, opened 2021-03-05
fix(h5p-server): correct generation of statement.object.id

### LINKED ISSUE #1147: Missing id in xAPI statements
The field statement.object.id is incorrect (undefined). image

Set via IIntegration.contents.url

sr258:
Closes #1147


sr258:
The UrlGenerator now also generates this URL.

I had to change the signature of IUrlGenerator for this, which is public. Should we consider it a break change even though its very small?

JPSchellenberg:
Well, since the last breaking chance did not trigger the major update we should use it as breaking change.

JPSchellenberg (approved):

PR #1210, opened 2021-03-22
feat(mongo-gridfs): add mongo-gridfs adapter

JPSchellenberg:
Hey,

I copied the mongos3-adapter and replaced s3 with gridfs. I still need to do the TemporaryFileStorage.
One thing that should be looked at is the `listFiles` command as I struggled to build a mongo-query that searchs for the beginning of the key. I tried `$text: {$serach: <string>}` and `{ $regex: <pattern> }`, but both failed to return the desired entries. Right now I query everything and filter the result with `.filter`, but I think that should be done in the query.

The tests can be run by
`DEBUG=h5p:* LOG_LEVEL=debug  npm run test:db:gridfs`


sr258:
I'm not sure if MongoGridFSLibraryStorage has much use, as library files will not go beyond 5 MB, which is the file size limit at which MongoLibraryStorage will stop working. We can offer it, but I don't think there are many use cases for it.

Do you want to move MongoLibraryStorage to the new package and remove it in h5p-mongo-s3? It makes sense, as you can avoid the aws-sdk dependency if you don't need it.

Will the classes work with Meteor's mongo? Meteor has its own Mongo driver, which manages subscriptions and might not be fully compatible.

I've fixed the listing files issue with a regexp find. I'm not convinced that this is the best way of doing it for large scales, as the regexp query is probably not indexed. I've thought of denormalizing the reference by adding a list of files to the content document. We can still do this later to optimize the class, though.

MongoGridFSContentStorage.getFileStream must implement the range option, as the interactive video editor and player won't work, if it doesn't.

sr258:
I've checked and I'm afraid the Meteor Mongo driver is too different to simply drop these classes in. I think there must be a specialized Meteor storage class.

You can check out my attempt at https://github.com/Lumieducation/h5p-meteor/tree/try-native-mongo-storage

The problem is that the Meteor mongo driver misses some function we use: `h5p:MongoGridFSContentStorage Error while listing all ids of content objects: cursor.toArray is not a function`.

JPSchellenberg:
I will work on the meteor-mongo driver in the other repo then.

PR #1343, opened 2021-04-25
refactor(scripts): added concurrently to improve speed

sr258:


sr258:
The idea here is to speed up development and CI pipelines by running scripts in parallel. I've moved the main build script to the postinstall script to get the correct order.

A caveat of doing this is that it's more difficult to see which script failed as all of the execute in parallel and you have to scroll a bit.

@JPSchellenberg what do you think, is it worth it?

PR #1362, opened 2021-04-30
fix(html-exporter): lazy loaded JS and CSS files work

sr258:


DegrangeM:
Just thinking, maybe we should encapsulate everything to avoid leaking `nativeSetLinkAttribute` to main scope ?

DegrangeM:
Not directly related to your PR, but about this line:

https://github.com/Lumieducation/H5P-Nodejs-library/blob/fix/export-lazy-loaded-files/packages/h5p-html-exporter/src/loadFileOverrides.js#L14

I think it should return the original value instead of undefined if the file is not found.

This make debugging harder and might have other bad side effect ?

DegrangeM:
Also https://github.com/Lumieducation/Lumi/issues/1483#issuecomment-829922481

Might be a good reason to also include .json file to the list of embeded files ? This will fix the timeline content type I think.

That way, we don't need to rewrite the ajax function (even if it's possible, note to myself if it's required one day : `nativeOpen = XMLHttpRequest.prototype.open; XMLHttpRequest.prototype.open = function(method, url, async, user, password){console.log('a',arguments);nativeOpen(method, url, async, user, password);}`).

sr258:
Thank you for your suggestions! I've encapsulated the code and also added a fallback for cases in which a file can't be found in the inlined resources.

So far, I've deliberately left out the library.json file (see https://github.com/lumieducation/h5p-nodejs-library/blob/17d194118ab3059fdba4c47bd75b733d95b60e56/packages/h5p-html-exporter/src/HtmlExporter.ts#L496), as they would add lots of nearly always unneccesary clutter. I try to keep the package as small as possible... 

And adding it to the bundle for Timeline isn't enough. (See [this](https://github.com/Lumieducation/H5P-Nodejs-library/tree/fix/export-timeline) branch.) Timeline seems to lazy load it's files in even another way, so the generated bundle still doesn't work. I don't think it's worth putting that much effort into a single content type if all others seem to work fine.

DegrangeM:
Yes you are right. In fact, the timeline library is a little old and the best way would be that one day, someone switch the timeline to the new version.

I will try to find the issue with timeline for 10min and if I don't find it I will giveup.

DegrangeM:
Actually, it was worth digging as the issue might happen one day in an other situation.

The issue come from the fact that the loaded file are called with a version parameter to force cache update when version change.

For exemple, it call for `TimelineJS-1.1/css/timeline.css?2.24`. Maybe remove ? and everything after when received file name ?

DegrangeM:
I tested with a fix (that is not with a good syntax (I just discovered that you can't put const inside if condition) but it was just a try) :
```
                let file = value.substr(12);
                if((paramindex = file.lastIndexOf('?')) !== -1) {
                   file = file.substr(0, paramindex); 
                }
```
and ... it work partially !

The timeline is display and is interactable, but there are two missing image file called from the css style. `loading.gif?v3.4 and timeline.png?v4.4 `).

We could also patch the img element (`HTMLImageElement`) but just adding the `?` fix is probably enough if we just want to fix the timeline content-type (but it might be a good fix for other content-type one day).

sr258:
I've added a regex to find these cases in one go and it seems to work. Having query parameters as a cache buster is quite normal, so we should add this behavior. The images still aren't loaded, but I wouldn't worry about this at this point...

There are still a few errors in the CI pipeline

sr258:
@DegrangeM Thanks for your work here, it was really helpful!

DegrangeM:
> @DegrangeM Thanks for your work here, it was really helpful!

You are welcome ! Thanks for fixing this !

PR #1685, opened 2021-08-19
fix(h5p-webcomponents): editor waits for initialization before saving

### LINKED ISSUE #1786: Error message "editorInstance of h5p editor not defined."
Users get the error message "editorInstance of h5p editor not defined." when loading H5Ps:
image

This is caused by the webcomponent not being fully initialized before the editor state is saved. I'm not certain why Lumi even tries to trigger the save, though.

sr258:
This might fix https://github.com/Lumieducation/Lumi/issues/1786, but as I can't reproduce the error, I'm not certain. It still won't hurt to delay saving until the initialization has finished.

### DIFF
--- a/packages/h5p-webcomponents/src/h5p-editor.ts
+++ b/packages/h5p-webcomponents/src/h5p-editor.ts
@@ -1,3 +1,4 @@
+import AwaitLock from 'await-lock';
 import type { IEditorModel, IContentMetadata } from '@lumieducation/h5p-server';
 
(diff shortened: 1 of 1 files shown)


sr258:
It's not the fix for the bug in https://github.com/Lumieducation/Lumi/issues/1786. Should we still merge this anyway @JPSchellenberg ? It's more clean theoretically, but I'm not sure if this race condition truly causes troubles in the real world.

PR #1775, opened 2021-09-19
Allow '@' character in the file name.

mujtabachang:
Added '@' character to the allowed list of filename characters.

US-ASCII characters are allowed for S3 bucket. Character '@' is a part of US-ASCII.

Reference: http://www.columbia.edu/kermit/ascii.html

H5P content indeed contains filenames with '@' characters. 

For example the file at:
[https://github.com/h5p/timelinejs/blob/stable/css/fancybox_sprite%402x.png](https://github.com/h5p/timelinejs/blob/stable/css/fancybox_sprite%402x.png)


sr258:
Thank you for the contribution!

PR #1853, opened 2021-10-31
refactor: enable go to definition between packages

sr258:


sr258:
This simplifies development as you can set breakpoints when running the debugger and you can jump between the packages.

PR #1871, opened 2021-11-07
fix(h5p-server): fix library install race conditions

### LINKED ISSUE #925: Improve library updates and installation
Currently library updates (1.2.0 -> 1.2.1) or installations (add H5P.Example-1.0) are not atomic, so it is possible that multiple requests try updating or installing libraries at the same time (this can happen even in single process setups due to the concurrency introduced when waiting for file IO). We should add a mechanism that introduces some kind of transactional behavior, e.g. by using redlock.

### LINKED ISSUE #1863: Concurrency of installing libraries leads to errors
If there are concurrent library installations, this can lead to problems. That's the reason why the CI integration tests have become flaky. (No idea, why this wasn't the case before). @JPSchellenberg 

The H5PEditor.saveH5P() - can save all real-world examples from the content-type-hub concurrently uploads and installs all 48 packages from the H5P Hub and this sometimes loads to breakages if there is a race condition in LibraryManager.installFromDirectory. (libraryExists has returned false in thread A and it continues to installLibrary; meanwhile thread B installs the library returns control to A and installLibrary fails for A)

We need a semaphore / lock that locks the library modification processes for a library using the ubername as key. This lock must support multi-process and multi-machine installations. We need a new interface for a locking mechanism that supports RedLock or some other locking mechanism.

sr258:
Introduces a new locking mechanism when installing libraries.

Before merging, I'd like to add a Redis implementation and make sure that the lock interface is generic enough and contains all necessary options.

Should fix the CI failures we've encountered recently. Closes #1863. Closes #925.


sr258:
@JPSchellenberg I consider the PR finished, you can review now.

JPSchellenberg:
The code looks good as far as I reviewed it (it's quite a lot :D) .
But as soon as I run the tests on my machine it says:
```
> @lumieducation/h5p-monorepo@7.0.1 test
> jest --maxWorkers=${BUILD_WORKERS-`nproc`}

sh: nproc: command not found
``` 

and then stops. Any hints if it is because of the missing `nproc`?  

I found that `sysctl -n hw.physicalcpu` should be used on macOS instead `nproc`: https://github.com/memkind/memkind/issues/33#issuecomment-316615308

sr258:
Yes, you need to create an alias to nproc on your mac, I believe: `alias nproc="sysctl -n hw.logicalcpu"`

JPSchellenberg:
Update: after a few minutes I got an error:
```
jest-haste-map: Watchman crawl failed. Retrying once with node crawler.
  Usually this happens when watchman isn't running. Create an empty `.watchmanconfig` file in your project's root folder or initialize a git or hg repository in your project.
  Error: Watchman error: query failed: synchronization failed: Operation timed out. Make sure watchman is running for this project. See https://facebook.github.io/watchman/docs/troubleshooting.
```
I will install watchman and see if it works.

sr258:
I've never seen watchman before. You can also remove the maxWorkers argument. I originally added it to make sure the runners in the CI aren't overtaxed, as Jest detect the cores of the actual CPU and not of the VM. On a dev machine it doesn't matter, as the physical cores is equivalent to the usable cores. The CI scripts don't call the NPM scripts any more, as the run better when they have the --runInBand argument.

JPSchellenberg:
I solved the issue: I had an old x86 installation of watchman which could not be run on the m1 mac. After uninstalling it it works perfectly fine. -  All tests pass on my machine.

JPSchellenberg (approved):
Just to make sure I understand what's going on:

When installing a library the LibraryManager now acquires a lock from the ILockProvider for a key derived from the Libraries UberName and if it succeeds the callback is invoked to install the new library, which has `installLibraryLockMaxOccupationTime` ms to install the library.

If the lock is not acquired that means that another instance is already installing the same library.

Is this correct? If so I haven't found any errors and the PR can be merged.

sr258:
Yes, that's how it works. The lock only fails if the lock acquisition timeout is reached. The effect is that library installation (of one library) is done sequentially and you don't get race conditions inside the library installation routine.

Theoretically it would also make sense to add the locks to the delete method, but as the delete method is also called from installLibrary when something goes wrong, this would lead to deadlocks.

PR #1886, opened 2021-11-12
feat(contentUserData): add feature to save and retrieve contentUserdata (closes #1014)

### LINKED ISSUE #1014: Implement the "save content state" option of H5P core
**Is your feature request related to a problem? Please describe.**
When using the node.js port of H5P core, you cannot save the current content's state and resume later.

**Describe the solution you'd like**
- I'd like the node.js port of H5P core to call instances' getCurrentState function (if present) in regular intervals and on instances' xAPI completed/progressed statements in order to save the instances' current state, cmp. https://github.com/h5p/h5p-php-library/blob/master/js/h5p.js#L222-L251
- I'd like the node.js port of H5P core to instantiate H5P content with previously saved states (if set) within the extras argument, cmp. https://github.com/h5p/h5p-php-library/blob/master/js/h5p.js#L929 and https://github.com/h5p/h5p-php-library/blob/master/js/h5p.js#L909 and https://github.com/h5p/h5p-php-library/blob/master/js/h5p.js#L109-L111 and https://github.com/h5p/h5p-php-library/blob/master/js/h5p.js#L97-L105

**Describe alternatives you've considered**
People who want the "save content state" option on their platform could come up with custom solutions to implement in content types, but that would not be efficient on a global scale and defeat the purpose of H5P content being shareable across platforms.

JPSchellenberg:
This PR will introduce the feature of saving and retreiving contentUserData.


JPSchellenberg:
Documentation and comments will follow tomorrow. 

Right now the contentUserData can be integrated directly into the Integration generated by the `H5PPlayer.generateIntegration`-method. Loading it via the GET request does not work. See https://github.com/Lumieducation/H5P-Nodejs-library/issues/1014#issuecomment-968139480

packages/h5p-server/src/types.ts
```diff
@@ -543,6 +544,48 @@ export interface IUser {
...
+    saveContentUserData(
+        contentId: ContentId,
+        dataType: string,
+        subContentId: string,
+        data: any,
+        user: IUser
+    ): Promise<void>;
+
+    /**
+     * Loads the contentUserData for given contentId and builds an array of IContentUserData which is used in the integtration object of the rendered H5P by the H5PPlayer.
```
sr258: This method should not be part of the Storage interface but in the Manager class, as it applies to all storage classes.

packages/h5p-server/src/types.ts
```diff
@@ -1533,6 +1576,13 @@ export interface IH5PConfig {
          */
         protocol?: 'http' | 'https';
     };
+    /*
+     * How often the user state of content is saved (in seconds). Set to false
+     * to disable saving user state. Note that the user state is only saved if
+     * the user object is passed into the render method of the player. You also
+     * must set ajax.contentUserData for state saving to work.
+     */
+    saveFreq: number | boolean;
```
sr258: I think the config option should be named in a a clearer way. The name Joubel chose is unclear and doesn't really help.

packages/h5p-server/src/H5PEditor.ts
```diff
@@ -97,7 +100,8 @@ export default class H5PEditor {
             'copyright-semantics': defaultCopyrightSemanticsLanguageFile
         }).t,
         private urlGenerator: IUrlGenerator = new UrlGenerator(config),
-        private options?: IH5PEditorOptions
+        private options?: IH5PEditorOptions,
+        public contentUserDataStorage?: IContentUserDataStorage
```
sr258, line 105: The options object should always be the last parameter. I had to add the ILockProvider parameter in the other PR and I put it into the objects. I think it would be best to refactor the constructor to have parameters like this storage: {libraryStorage: ILibraryStorage, contentStorage: IContentStorage, ...}. That way we avoid too many breaking changes when adding new optional parameters.

packages/h5p-server/src/H5PEditor.ts
```diff
@@ -126,6 +130,9 @@ export default class H5PEditor {
             temporaryStorage,
             this.config
         );
+        this.contentUserDataManager = new ContentUserDataManager(
+            contentUserDataStorage
```
sr258, line 144: What happens when contentUserDataStorage is undefined?

packages/h5p-server/test/H5PPlayer.renderHtmlPage.test.ts
```diff
@@ -39,7 +39,7 @@ describe('Rendering the HTML page', () => {
                     <script>
                         window.H5PIntegration = {
                   "ajax":{
-                     "contentUserData":"/h5p/contentUserData",
+                     "contentUserData":"/h5p/contentUserData/:contentId/:dataType/:subContentId",
```
sr258: We should consider moving to Jest Snapshots. Mainting this test file is quite annoying. I have touched it myself several times and it's very redundant work.

package.json
```diff
@@ -36,6 +36,7 @@
         "semantic-release": "semantic-release",
         "start:dbs": "docker-compose -f scripts/mongo-s3-docker-compose.yml up -d",
         "start": "npm run start --prefix packages/h5p-examples",
+        "start:rest": "npm run start --prefix packages/h5p-rest-example-server",
```
sr258: It only starts the server, the name should make it clear.

sr258:
I've added a few first remarks.

packages/h5p-server/src/types.ts
JPSchellenberg: I agree. Any suggestions? I would suggest `contentUserDataSaveInterval` or `contentUserStateSaveInterval`.

packages/h5p-server/src/H5PEditor.ts
JPSchellenberg: Yeah. I put it last to avoid a breaking change since we basically only introduce a feature. I thought about the storage object as well and think that would be the best way to go.

packages/h5p-server/src/H5PEditor.ts
JPSchellenberg: The contentUserDataManger is constructed, but if the methods are called they return `undefined`:

https://github.com/Lumieducation/H5P-Nodejs-library/pull/1886/files#diff-175df6189e4a5c5418312fab8fcb2634717e3565726190172a885f90a9fd26e7R73

Should I use another way of doing this?

packages/h5p-server/test/H5PPlayer.renderHtmlPage.test.ts
JPSchellenberg: Ok, I made an issue for that: https://github.com/Lumieducation/H5P-Nodejs-library/issues/1892

packages/h5p-server/src/types.ts
sr258: I'd opt for contentUserStateSaveInterval.

packages/h5p-server/src/H5PEditor.ts
sr258: I see. I'd have implemented this in H5PEditor, but it can also be done in ContentUserDataManager

packages/h5p-server/src/types.ts
JPSchellenberg: The problem is that if it is only in the Manager class you have no idea/indication how many subContentIds there are.

Right now the implementation of `generateContentUserDataIntegration` the `IContentUserDataStorage` has to load all ContentUserData and build an array where the position in the array is equal to the subContentId.

If you use `loadContentUserData` from the Manager you don't know how many subContentIds / dataTypes there are. 

Or am I missing something?

sr258: In this case the interface should include a method like listByContent(contentId, userId) and the actual integration construction should be done in the Manager.

packages/h5p-express/src/ContentUserDataRouter/ContentUserDataController.ts
```diff
@@ -0,0 +1,59 @@
...
+     * Returns the userState for given contentId, dataType and subContentId
+     */
+    public getContentUserData = async (
+        req: IRequestWithUser,
+        res: express.Response<{
+            data: string;
+            success: boolean;
+        }>
+    ): Promise<void> => {
+        const { contentId, dataType, subContentId } = req.params;
```
sr258, line 34: potentially unsafe parameters must be validated / sanitized

packages/h5p-express/src/ContentUserDataRouter/ContentUserDataController.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Saving a userState for given contentId, datType and subContentId
+     *
+     */
+    public postContentUserData = async (
+        req: IPostContentUserDataRequest,
+        res: express.Response
+    ): Promise<void> => {
+        const { contentId, dataType, subContentId } = req.params;
```
sr258, line 61: potentially unsafe parameters must be validated / sanitized

packages/h5p-server/src/ContentManager.ts
```diff
@@ -21,10 +22,15 @@ const log = new Logger('ContentManager');
...
+     * @param contentUserDataStorage The contentUserDataStorage to delete contentUserData for content when it is deleted
      */
-    constructor(public contentStorage: IContentStorage) {
+    constructor(
+        public contentStorage: IContentStorage,
+        public contentUserDataStorage?: IContentUserDataStorage
+    ) {
         log.info('initialize');
         this.contentStorage = contentStorage;
+        this.contentUserDataStorage = contentUserDataStorage;
```
sr258: this.contentUserDataStorage is already set as it is set by the public constructor argument

packages/h5p-server/src/ContentManager.ts
```diff
@@ -90,14 +96,27 @@ export default class ContentManager {
...
-     * Deletes a piece of content and all files dependent on it.
+     * Deletes a piece of content, the corresponding contentUserData and all files dependent on it.
      * @param contentId the piece of content to delete
      * @param user the user who wants to delete it
      */
     public async deleteContent(
         contentId: ContentId,
         user: IUser
     ): Promise<void> {
+        if (this.contentUserDataStorage) {
```
sr258, line 107: It might be sensible to acquire a lock that is also acquired when the user posts the content state. Otherwise a race condition is possible in which the content state is deleted, created again and then the content is deleted. The effect would be that there are orphaned content states. Alternatively it you could delete the content state after deleting the content.

packages/h5p-server/src/ContentManager.ts
```diff
@@ -90,14 +96,27 @@ export default class ContentManager {
...
+            try {
+                await this.contentUserDataStorage.deleteContentUserData(
+                    contentId,
+                    user
```
sr258: The content user data must be deleted for all users, not just the current.

packages/h5p-server/src/ContentUserDataManager.ts
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * The ContentUserDataManager takes care of saving user data and states. It only contains storage-agnostic functionality and
+ * depends on a ContentUserDataStorage object to do the actual persistence.
+ */
+export default class ContentUserDataManager {
+    /**
+     * @param contentUserDataStorage The storage object
+     */
+    constructor(contentUserDataStorage: IContentUserDataStorage) {
```
sr258: Using a private constructor argument eliminated the need to assign the property and even declare it.

packages/h5p-server/src/ContentUserDataManager.ts
```diff
@@ -0,0 +1,154 @@
...
+        contentId: ContentId,
+        dataType: string,
+        subContentId: string,
+        user: IUser
+    ): Promise<string> {
+        if (!this.contentUserDataStorage) {
+            return undefined;
+        }
+
+        if (!this.contentUserDataStorage.loadContentUserData) {
```
sr258: Is this really necessary? An implementation wouldn't make much sense without this method, as it's not optional. We don't do this is the library anywhere else.

packages/h5p-server/src/ContentUserDataManager.ts
```diff
@@ -0,0 +1,154 @@
...
+        return;
+    }
+
+    /**
+     * Loads the contentUserData for given contentId, dataType and subContentId
+     * @param contentId The id of the content to load user data from
+     * @param dataType Used by the h5p.js client
+     * @param subContentId The id provided by the h5p.js client call
+     * @param user The user who owns this object
+     * @returns the saved state as string
```
sr258: or undefined when not found

packages/h5p-server/src/ContentUserDataManager.ts
```diff
@@ -0,0 +1,154 @@
...
+     * @param dataType Used by the h5p.js client
+     * @param subContentId The id provided by the h5p.js client call
+     * @param userState The userState as string
+     * @param user The user who owns this object
+     * @returns the saved state as string
+     */
+    public async saveContentUserData(
+        contentId: string,
+        dataType: string,
+        subcontentId: string,
```
sr258: mismatch with JSDocs name

packages/h5p-server/src/ContentUserDataManager.ts
```diff
@@ -0,0 +1,154 @@
...
+        userState: string,
+        invalidate: 0 | 1,
```
sr258: boolean would be better here; the library doesn't use 0 or 1 for booleans anywhere else, except if it passes values to the client

packages/h5p-server/src/index.ts
```diff
@@ -85,6 +88,7 @@ export {
     streamToString,
     ContentFileScanner,
     ContentTypeCache,
+    ContentUserDataManager,
```
sr258, line 97: Is it necessary to export it? The other managers are exported as they are used by the html exporter package, but I'm not sure if anyone might want to use the ContentUserDataManager outside of the library. If we export too many classes, it becomes more difficult to change signatures.

packages/h5p-server/test/ContentManager.test.ts
```diff
@@ -129,6 +131,33 @@ describe('ContentManager', () => {
...
+                const contentId = await contentManager.createOrUpdateContent(
+                    mockupMetadata,
+                    mockupParameters,
+                    user
+                );
+
+                await expect(
+                    contentManager.deleteContent(contentId, user)
+                ).resolves.toBeUndefined();
+                await expect(
```
sr258: It would be better if the test created content states for several users and checks if the content is really deleted for all users.

packages/h5p-server/test/ContentUserDataManager.test.ts
```diff
@@ -0,0 +1,211 @@
+import ContentUserDataManager from '../src/ContentUserDataManager';
+
+import MockContentUserDataStorage from './__mocks__/ContentUserDataStorage';
+
+import User from './User';
+
+describe('ContentUserDataManager', () => {
```
sr258, line 7: The tests here mostly check if function of the storage class are called. I think it would be more useful if tests checked actual functionality, like saving a state and retrieving it later. Just checking if calls are passed to the storage class might lead to a false sense of secruity as it produces good line coverage, but doesn't check any real functionality or logic.

packages/h5p-server/src/H5PEditor.ts
```diff
@@ -1112,7 +1123,7 @@ export default class H5PEditor {
             },
             libraryConfig: this.config.libraryConfig,
             postUserStatistics: false,
-            saveFreq: false,
+            saveFreq: this.config.contentUserStateSaveInterval,
```
sr258: I guess the editor also needs the url where it can save states

JPSchellenberg: Yep. @otacke mentioned that the editor uses it to store data about tours.

packages/h5p-server/test/ContentUserDataManager.test.ts
JPSchellenberg: I agree with that, but we should think about separating two tests:
unit-test which solely tests the "unit"  and integration tests. If you add an actual storage class you also test the functionality of the storage class. If you want a pure unit test you have to mock everything else.
I will add an integration test as a separate test.

packages/h5p-server/src/index.ts
JPSchellenberg: If you want to allow users to integrate it under custom routes then yes. It might be necessary to put custom logic into a route and therefore it would be good if you can import the manager and handle the state saving on your own.

packages/h5p-server/src/ContentUserDataManager.ts
JPSchellenberg: I know, but it is what the core sends. But I guess changing it to boolean would not be a problem.

sr258: The paramters of the core are already changed in other places. We simply have to map the values to the internal types and back. The core is very inconsistent in how booleans are used (0, "0" or false are used) and it's better to have consistence within the library.

packages/h5p-server/src/ContentUserDataManager.ts
JPSchellenberg: ok, I will take it out.

packages/h5p-server/src/ContentManager.ts
JPSchellenberg: contentUserData is deleted based on the contentId - the user is only passed to check if he has permissions to delete the states.

packages/h5p-server/src/ContentManager.ts
JPSchellenberg: Ok, I will delete it after the content is deleted.

packages/h5p-server/test/ContentManager.test.ts
JPSchellenberg: The contentUserData is deleted based on the contentId. I will add a separate function for a specific user.

packages/h5p-express/src/ContentUserDataRouter/ContentUserDataController.ts
JPSchellenberg: I will sanitize these things in the ContentUserDataManager as there will be no need for the user to do so if the user decides to handle the implementation on their own.

docs/usage/h5p-editor-constructor.md
```diff
@@ -114,3 +114,13 @@ for an implementation sample using the urlGenerator.
 
 Allows you to customize styles and scripts of the client. Also allows passing in
 a lock implementation (needed for multi-process or clustered setups).
+
+## contentUserDataStorage (optional)
+
+The `contentUserDataStorage` handles saving and loading user states, so users can continue where they left off when they reload the page or come back later. It must implement the `IContentUserDataStorage` interface. You can find a reference implementation in [`/packages/h5p-examples/src/contentUserDataStorage.ts`](/packages/h5p-examples/src/contentUserDataStorage.ts):
+nder `./h5p/content` you can use the sample
```
sr258: something is cut off here

packages/h5p-examples/src/contentUserDataStorage.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    public deleteContentUserDataByUserId(
+        contentId: ContentId,
+        userId: string,
+        requestingUser: IUser
+    ): Promise<void> {
+        userData = userData.filter(
+            (data) => data.contentId !== contentId && data.userId !== userId
+        );
+        return;
```
sr258: redundant; no Promise returned

packages/h5p-examples/src/contentUserDataStorage.ts
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    public deleteAllContentUserDataByContentId(
+        contentId: ContentId,
+        requestingUser: IUser
+    ): Promise<void> {
+        userData = userData.filter((data) => data.contentId !== contentId);
+        return;
```
sr258: redundant; no Promise returned

packages/h5p-examples/src/contentUserDataStorage.ts
```diff
@@ -0,0 +1,135 @@
...
+            contentId,
+            dataType,
+            subContentId,
+            userState,
+            // invalidate,
+            // preload,
+            userId: user.id
+        });
+
+        return;
```
sr258: redundant

packages/h5p-examples/src/contentUserDataStorage.ts
```diff
@@ -0,0 +1,135 @@
...
+                (data) =>
+                    data.contentId === contentId &&
+                    data.dataType === dataType &&
+                    data.subContentId === subContentId &&
+                    data.userId === user.id
+            )[0]?.userState
+        );
+    }
+
+    public saveContentUserData(
```
sr258: Does this method return a Promise if it's not async?

packages/h5p-examples/src/contentUserDataStorage.ts
```diff
@@ -0,0 +1,135 @@
...
+    > {
+        return Promise.resolve(
+            userData.filter(
+                (data) => data.contentId === contentId && data.userId === userId
+            )
+        );
+    }
+}
+
+export default new ContentUserDataStorage();
```
sr258: The library doesn't use singletons and they are typically considered an anti-pattern. The code that uses the class should construct its own instance with new.

packages/h5p-examples/src/createH5PEditor.ts
```diff
@@ -8,6 +8,7 @@ import * as dbImplementations from '@lumieducation/h5p-mongos3';
 import RedisLockProvider from '@lumieducation/h5p-redis-lock';
 import { ILockProvider } from '@lumieducation/h5p-server';
 
+import contentUserDaraStorage from './contentUserDataStorage';
```
sr258: Should only import the class and construct it's own object

packages/h5p-examples/src/express.ts
```diff
@@ -103,7 +106,9 @@ const start = async (): Promise<void> => {
         config,
         undefined,
         undefined,
-        (key, language) => translationFunction(key, { lng: language })
+        (key, language) => translationFunction(key, { lng: language }),
+        undefined,
+        contentUserDataStorage
```
sr258: Should get the storage from the editor like libraryStorage

packages/h5p-express/src/ContentUserDataRouter/ContentUserDataController.ts
```diff
@@ -0,0 +1,89 @@
+import * as express from 'express';
+
```
sr258: I've separated import groups by external imports and internal ones.

packages/h5p-express/src/ContentUserDataRouter/ContentUserDataController.ts
```diff
@@ -0,0 +1,89 @@
...
+        if (!userState) {
+            res.status(404).json({ data: undefined, success: false });
+        } else {
+            res.status(200).json({ data: userState, success: true });
+        }
+    };
+
+    /**
+     * Saving a userState for given contentId, datType and subContentId
```
sr258: dat_a_type

packages/h5p-express/src/ContentUserDataRouter/ContentUserDataExpressRouter.ts
```diff
@@ -0,0 +1,62 @@
+import { Router } from 'express';
+
```
sr258, line 3: import groups

packages/h5p-express/src/ContentUserDataRouter/ContentUserDataExpressRouter.ts
```diff
@@ -0,0 +1,62 @@
...
+import {
+    errorHandler,
+    undefinedOrTrue,
+    catchAndPassOnErrors
+} from '../expressErrorHandler';
+
+import ContentUserDataController from './ContentUserDataController';
+/**
+ * This router implements necessary routes for the contentUserData (userState) to work.
+ * If you only want certain routes, you can specify this in the options parameter.
```
sr258: JSDocs doesn't match signature

packages/h5p-express/src/ContentUserDataRouter/ContentUserDataExpressRouter.ts
```diff
@@ -0,0 +1,62 @@
...
+    options: { handleErrors: boolean } = { handleErrors: true },
+    languageOverride: string | 'auto' = 'auto'
+): Router {
+    const router = Router();
+
+    const contentUserDataController = new ContentUserDataController(
+        contentUserDataManager
+    );
+
+    router.get(
```
sr258, line 35: The router doesn't use the config; I'm not sure if this makes sense, as it's hardcoded by the H5P client anyway; maybe the config shouldn't include the placeholders then

packages/h5p-express/src/ContentUserDataRouter/ContentUserDataExpressRouter.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    router.post(
+        `/:contentId/:dataType/:subContentId`,
+        catchAndPassOnErrors(
+            contentUserDataController.postContentUserData,
+            undefinedOrTrue(options?.handleErrors)
+        )
+    );
+
+    router.post(
```
sr258: Should be configured through setFinishedUrl

packages/h5p-express/test/ContentUserDataRouter.test.ts
```diff
@@ -0,0 +1,154 @@
...
+            contentId,
+            dataType,
+            subContentId,
+            body.data,
+            false,
+            true,
+            user
+        );
+        expect(res.status).toBe(200);
+        // expect(res.body).toBe({ data: mockReturnData, success: true });
```
sr258: obsolete?

packages/h5p-rest-example-server/config.json
```diff
@@ -6,6 +6,7 @@
     "sendUsageStatistics": false,
     "hubRegistrationEndpoint": "https://api.h5p.org/v1/sites",
     "hubContentTypesEndpoint": "https://api.h5p.org/v1/content-types/",
+    "contentUserDataUrl": "/contentUserData/:contentId/:dataType/:subContentId",
```
sr258: It would be good if the rest example also included the user state feature, as I think it is used a lot.

packages/h5p-server/src/implementation/H5PConfig.ts
```diff
@@ -98,8 +99,9 @@ export default class H5PConfig implements IH5PConfig {
         protocol?: 'http' | 'https';
     };
     public sendUsageStatistics: boolean = false;
-    public setFinishedUrl: string = '/setFinished';
+    public setFinishedUrl: string = '/contentUserData/setFinished';
     public siteType: 'local' | 'network' | 'internet' = 'local';
+    public contentUserStateSaveInterval: number = 30; // the interval to save the contentUserData in seconds
```
sr258: I think it would make sense to change this into milliseconds and convert the interval into seconds for H5P. That way the configuration isn't inconsistent.

packages/h5p-server/src/ContentManager.ts
```diff
@@ -90,15 +94,31 @@ export default class ContentManager {
...
+                );
+            } catch (error) {
+                log.error(
+                    `could not delete contentUserData with contentId ${contentId}`
+                );
+                log.error(error);
+            }
+        }
+
+        return;
```
sr258: redundant

packages/h5p-server/src/ContentUserDataManager.ts
```diff
@@ -0,0 +1,211 @@
+import {
```
sr258, line 1: internal import group comes after external imports

packages/h5p-server/src/ContentUserDataManager.ts
```diff
@@ -0,0 +1,211 @@
...
+     * @param userId the userId for which the contentUserData object should be deleted
+     * @param requestingUser The user who wants to delete the content (not the user the contentUserData belongs to)
+     */
+    public async deleteContentUserDataByUserId(
+        contentId: ContentId,
+        userId: string,
+        requestingUser: IUser
+    ): Promise<void> {
+        if (this.contentUserDataStorage) {
+            log.info(
```
sr258: Should be debug

packages/h5p-server/src/ContentUserDataManager.ts
```diff
@@ -0,0 +1,211 @@
...
+                `deleting contentUserData for ContentId ${contentId} and userId ${userId}`
+            );
+            return this.contentUserDataStorage.deleteContentUserDataByUserId(
+                contentId,
+                userId,
+                requestingUser
+            );
+        }
+        return;
```
sr258: redundant

packages/h5p-server/src/ContentUserDataManager.ts
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    public async deleteAllContentUserDataByContentId(
+        contentId: ContentId,
+        requestingUser: IUser
+    ): Promise<void> {
+        if (this.contentUserDataStorage) {
+            log.info(`deleting all contentUserData for ContentId ${contentId}`);
```
sr258: Should be debug

packages/h5p-server/src/ContentUserDataManager.ts
```diff
@@ -0,0 +1,211 @@
...
+            return this.contentUserDataStorage.deleteAllContentUserDataByContentId(
+                contentId,
+                requestingUser
+            );
+        }
+        return;
```
sr258: redundant

packages/h5p-server/src/ContentUserDataManager.ts
```diff
@@ -0,0 +1,211 @@
...
+        contentId: ContentId,
+        dataType: string,
+        subContentId: string,
+        user: IUser
+    ): Promise<string> {
+        if (!this.contentUserDataStorage) {
+            return undefined;
+        }
+
+        log.info(
```
sr258: should be debug (and also log calls below)

packages/h5p-server/src/ContentUserDataManager.ts
```diff
@@ -0,0 +1,211 @@
...
+            return undefined;
+        }
+
+        const states = await this.contentUserDataStorage.listByContent(
+            contentId,
+            user.id
+        );
+
+        const sortedStates = states.sort(
+            (a, b) => Number(a.subContentId) - Number(b.subContentId)
```
sr258, line 132: Is this a save conversion? Is the subcontent id always a number?

packages/h5p-server/src/ContentUserDataManager.ts
```diff
@@ -0,0 +1,211 @@
...
+
+        const mappedStates = sortedStates.map((state) => ({
+            [state.dataType]: state.userState
+        }));
+
+        return mappedStates;
+    }
+
+    /**
+     * Saves data when a user completes content. T
```
sr258: T seems like a leftover

packages/h5p-server/src/ContentUserDataManager.ts
```diff
@@ -0,0 +1,211 @@
...
+     * @param contentId The content id to delete.
+     * @param score the score the user reached as an integer
+     * @param maxScore the maximal score of the content
```
sr258: maximum

packages/h5p-server/src/ContentUserDataManager.ts
```diff
@@ -0,0 +1,211 @@
...
+    ): Promise<void> {
+        log.info(
+            `saving contentUserData for user with id ${user.id} and contentId ${contentId}`
+        );
+
+        const sanitizedUserState = sanitizeHtml(userState);
+
+        if (!sanitizedUserState || sanitizedUserState === '') {
+            log.error(`no userState provided for ${contentId}`);
+            throw new Error('no-userState');
```
sr258: Is this a message that should be sent to the user? There's the H5PError class for that. The convention for errors used in the rest of the library is 'no-user-state'

packages/h5p-server/src/ContentUserDataManager.ts
```diff
@@ -0,0 +1,211 @@
...
+import sanitizeHtml from 'sanitize-html';
+import Logger from './helpers/Logger';
+
+const log = new Logger('ContentUserDataManager');
+
+/**
+ * The ContentUserDataManager takes care of saving user data and states. It only contains storage-agnostic functionality and
+ * depends on a ContentUserDataStorage object to do the actual persistence.
+ */
+export default class ContentUserDataManager {
```
sr258, line 17: The finished state should also be retreived somehow: all states for the contentId and the individual state for a single user

packages/h5p-server/src/H5PPlayer.ts
```diff
@@ -60,7 +62,8 @@ export default class H5PPlayer {
             // English if the implementation does not pass us a proper one.
             client: englishClientStrings
         }).t,
-        private options?: IH5PPlayerOptions
+        private options?: IH5PPlayerOptions,
+        contentUserDataStorage?: IContentUserDataStorage
```
sr258, line 67: The options should come last; I would opt for a breaking change and collect all storage classes in a separate object

packages/h5p-server/src/H5PEditor.ts
sr258: The storage object is still a good idea.

packages/h5p-server/src/types.ts
```diff
@@ -543,6 +544,112 @@ export interface IUser {
     type: 'local' | string;
 }
 
+/**
+ * Implementations can implement the IContentUserDataStorage interface and pass it to the constructor of
+ * H5PEditor or H5PPlayer if they wish to keep the user state and allows users to continue where they left off.
+ */
+export interface IContentUserDataStorage {
```
sr258, line 623: The interface should also include capbilities to delete the user data for a specific user only across all content objects and to list all user data for a specific user. (GDPR compliance)

packages/h5p-server/src/ContentUserDataManager.ts
```diff
@@ -0,0 +1,211 @@
...
+        userState: string,
+        invalidate: boolean,
+        preload: boolean,
+        user: IUser
+    ): Promise<void> {
+        log.info(
+            `saving contentUserData for user with id ${user.id} and contentId ${contentId}`
+        );
+
+        const sanitizedUserState = sanitizeHtml(userState);
```
sr258: Does this work for JSON object? It's JSON string after all...

sr258:
I haven't tested it live yet, only looked at the code. I think there should also be a file storage class that keeps user states on the disk for simple implementations (or future uses in Lumi Player). There's also need for a MongoDB class.

sr258 (changes requested):

packages/h5p-server/src/H5PEditor.ts
JPSchellenberg: But I would do that in another PR as it is not really related to this feature. We shouldn't do too many things in one PR.

packages/h5p-server/src/ContentUserDataManager.ts
JPSchellenberg: As far as I reconstructed it the subContentId equals the index in the array that is in [contentUserData](https://github.com/Lumieducation/H5P-Nodejs-library/blob/feat/1014-contentUserData/packages/h5p-server/src/H5PPlayer.ts#L391) field in the Integration object. Therefore it should always be a number.

packages/h5p-server/src/types.ts
JPSchellenberg: But from where in the library should it be called?

packages/h5p-server/src/H5PPlayer.ts
JPSchellenberg: I will make a separate PR for that when this PR is merged.

packages/h5p-server/src/ContentUserDataManager.ts
JPSchellenberg: For what would you sanitise the object? The sanitizeHtml removes all javascript code, but is there something I can't think of?

packages/h5p-server/src/types.ts
sr258: The implementation can do it on its own. It constructs the storage, so it can also call the methods.

packages/h5p-server/src/ContentUserDataManager.ts
sr258: I wouldn‘t sanitize it at all. It‘s inherently unsafe, but we can’t do anything about it. The other args might need validation, but I wouldn‘t do it for the arbitrary data in the actual state.

packages/h5p-server/src/ContentUserDataManager.ts
JPSchellenberg: But wouldn't that be some kind of round-trip? The user would have to implement a corresponding method in the ContentUserDataStorage. I can't think of a scenario where a user would use a method on the manager class to retrieve data that way or am I missing something?

packages/h5p-server/src/ContentUserDataManager.ts
JPSchellenberg: Alright. How would validate the other arguments? Just by checking the typeof?

packages/h5p-server/src/ContentUserDataManager.ts
sr258: Yes, it can be left out in the manager, but should be included in the interface.

packages/h5p-server/src/ContentUserDataManager.ts
sr258: Yes, that‘ll do. I‘d check the number and boolean types in particular. These tend to be inconsistent in H5P.

packages/h5p-express/src/ContentUserDataRouter/ContentUserDataExpressRouter.ts
JPSchellenberg: How do I check this again?

packages/h5p-express/src/ContentUserDataRouter/ContentUserDataExpressRouter.ts
JPSchellenberg: You are right.. I will remove the placeholders

packages/h5p-express/src/ContentUserDataRouter/ContentUserDataExpressRouter.ts
sr258: It's only a divergence between the /** **/ part and the signature.

sr258:
@JPSchellenberg What's the status of this PR? It would make sense to merge it in soon to avoid having a long running PR that blocks further work.

JPSchellenberg:
I try to finish this PR as soon as possible. So far the only things left on my todo-list are:
- [x] include in rest example
- [x] router path should be configurable through setFinishedUrl

Do you have any additions?

sr258:
@JPSchellenberg I've reviewed the PR in depth and there is one issue that must be resolved before it can be merged into master and published:

The `invalidate` and `preload` flags have some meaning that is not used in the implementation. 
  - invalidate: The system should remove user content when content is updated
  - preload: Content state should only be added to the integration if this flag is set
Because of the meaning of these flags the IContentUserDataStorage has to be changed, too. The invalidate and preload flags must be returned in loadContentUserData and the delete methods should have the ability to differentiate between `invalidate` on and off

I believe that there should also also be a way to get the data of saveFinished through the interface, e.g. through an optional method. This can also be extended later, as it doesn't break things.

Also not a merge blocker but something needed for a release: we need a file storage implementation and a Mongo implementation that implementations can use. I don't believe there are many people around who don't use our default storage implementations, so this is a huge hurdle to using this feature.

JPSchellenberg:
@sr258  I included the rest-example and the invalidate and preload flags.
But for the invalidate I am not sure how to do it properly. Should the contentUserData be removed when invalidate is set to 1? Or should it be updated with an "invalidate-flag"?

sr258:
I haven't checked in the PHP code (which would be the best place to look for the exact behavior of the flag), but I think it should work like this:

- always replace the old state when new one is saved (regardless of `invalidate` flag); it might make sense if there was an update operation in the storage class for this (could be more performant in some situations than a delete and update operation; avoids race conditions)
- if state has the `invalidate` flag, it should be deleted if the *content* is updated by the content author (if the author changes the number of blanks in a fill-in-the-blanks exercise, the old state is incompatible with the new one)

I wonder what the PHP implementation does if a user has the H5P open and the author changes the content at the same time. It would lead to an invalid state once the user reloads it... You would probably need to know which revision of a piece of content the state belongs to. I'm not sure if this is implemented in PHP. You get a message in the client (not sure if it comes from the plugin or the core, though) that notifies the user that the state had to be deleted as the content was upgraded. I think looking at the PHP implementation will answer this.

JPSchellenberg:
I looked at the php code and it is indeed like you described. ContentUserData with the invalid-field set to true is deleted when the content changes: https://github.com/h5p/h5p-php-library/blob/d3d003e978ae361bcc5d2f0f36bb349923e61636/h5p.classes.php#L2161

I implemented that behaviour into the library.

PR #1954, opened 2021-12-03
feat(theme): add themes

JPSchellenberg:


packages/h5p-server/src/renderers/theme.ts
```diff
@@ -0,0 +1,772 @@
+import { ITheme } from '../types';
```
sr258, line 1: The stylesheet replicates a lot of the CSS properties of the styles. I think it should only include overrides and avoid duplicating CSS properties that are already set by H5P.

I still think that it would be sensible to use SCSS here.

packages/h5p-server/src/renderers/player.ts
```diff
@@ -11,7 +12,7 @@ export default (model: IPlayerModel): string => `<!doctype html>
     ${model.scripts
         .map((script) => `<script src="${script}"></script>`)
         .join('\n    ')}
-
+    <style>${generateThemeCSS(model.theme)}</style>
```
sr258: The theme should be optional

packages/h5p-examples/config.json
```diff
@@ -15,5 +15,16 @@
         "H5P.CoursePresentation": ["H5P.MathDisplay"],
         "H5P.InteractiveVideo": ["H5P.MathDisplay"],
         "H5P.DragQuestion": ["H5P.MathDisplay"]
+    },
+    "theme": {
```
sr258, line 21: Further customization options could be:

- corner rounding other button decorations
- font sizes and font faces
- margins and paddings

packages/h5p-server/src/H5PEditor.ts
```diff
@@ -155,8 +157,10 @@ export default class H5PEditor {
             );
         }
 
-        this.globalCustomStyles =
-            this.options?.customization?.global?.styles || [];
+        this.globalCustomStyles = this.options?.customization?.global
+            ?.styles || [
```
sr258: The theme should not replace the global styles; they should be concatenated I guess

packages/h5p-server/src/renderers/theme.ts
sr258: It would also allow scenarios in which admins change the override SCSS and don't have to touch code.

packages/h5p-server/src/types.ts
```diff
@@ -2089,3 +2092,16 @@ export interface ILockProvider {
         options: { timeout: number; maxOccupationTime: number }
     ): Promise<T>;
 }
+
+export interface ITheme {
+    themeUrl: string;
+    fontColor: string;
+    disabledFontColor: string;
+    backgroundColor: string;
+    paperBackgroundColor: string;
```
sr258: "paper" is a Material UI term; is there a more generic term?

sr258:
It would also make sense to accomodate scenarios in which you have a per-content object theme that is configurable by the content author.

packages/h5p-examples/config.json
JPSchellenberg: I already planed to add those.

packages/h5p-server/src/types.ts
JPSchellenberg: secondaryBackground?

packages/h5p-server/src/renderers/player.ts
JPSchellenberg: It is optional. When the theme is not defined it returns an empty string.

packages/h5p-server/src/H5PEditor.ts
JPSchellenberg: It is not replaced but entered as the first entry in the array .

sr258: Really? From what I see it takes the global styles and if it is undefined it now takes the theme. They are not combined. The && doesn‘t work here, as this only works in React.

packages/h5p-server/src/types.ts
sr258: Ok

packages/h5p-server/src/renderers/player.ts
sr258: On second thought I think the way to inject the theme with the global custom style is better, as it doesn‘t require changes in the client and it will also work with the iframe embed type.

packages/h5p-express/src/ThemeRouter/ThemeExpressRouter.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { H5PEditor } from '@lumieducation/h5p-server';
+import {
+    errorHandler,
+    undefinedOrTrue,
+    catchAndPassOnErrors
+} from '../expressErrorHandler';
+
+/**
+ * This router implements all Ajax calls necessary for the H5P (editor) client to work.
```
sr258: Description is not correct

packages/h5p-server/src/types.ts
```diff
@@ -2089,3 +2092,24 @@ export interface ILockProvider {
         options: { timeout: number; maxOccupationTime: number }
     ): Promise<T>;
 }
+
+export interface ITheme {
+    themeUrl: string;
```
sr258: This property is used nowhere else? What is its function?

sr258:
There are still details in content type editors that aren't styled: (didn't look at all of them yet)

Bildschirmaufnahme_Auswahlbereich_20220226182608
Bildschirmaufnahme_Auswahlbereich_20220226182621
Bildschirmaufnahme_Auswahlbereich_20220226182634
Bildschirmaufnahme_Auswahlbereich_20220226182700
Bildschirmaufnahme_Auswahlbereich_20220226182715
Bildschirmaufnahme_Auswahlbereich_20220226182750
Bildschirmaufnahme_Auswahlbereich_20220226182804
Bildschirmaufnahme_Auswahlbereich_20220226182817
Bildschirmaufnahme_Auswahlbereich_20220226182914
Bildschirmaufnahme_Auswahlbereich_20220226182948

sr258:
I've fixed most of the issues above in the editor. There's mostly work to be done in the course presentation editor (it's various submenus).

I've noticed that it seems like a bad idea to override the `color` property of the `body` as this has many side effects. H5P is so complex that it's difficult to predict where there might be a panel with a hardcoded background color, so setting the `color` globally is dangerous as it can easily lead to situations in which there's white on white. It would probably be better to set the color in a more restricted way.

I haven't really looked at the player yet.

sr258:
I've reviewed the state of the theme in the player and I think the approach of the PR isn't working out. Nearly all content types are broken in some way. I've reviewed about 40% (results below)

I suggest this:

- We have a list of content types that are supported (separate for editor and player).
- There are two themes, one for editor, one for player
- The CSS of the theme is never global but always scoped to the content type (in the player, and as far as possible also in the editor)
- The H5PEditor and H5PPlayer object decide whether to add a theme to the UI based on the content type
- There's a callback that implementations can pass in to disable themes for content (so that it is a setting for the content object)
- We go through the list of content types one by one and write SCSS for it, checking every feature and option. We avoid duplication using SCSS imports and mixins. 

@JPSchellenberg @jankapunkt What do you think of this approach?
Cornell 1
Cornell 2
Dictation
Dictation 2
Drag and Drop
Essay 1
Essay 2
Fill in the Blanks
Find the hotspot
Find the words
Guess the answer
Image Choice
Image Pairing
Image Sequence
image slider
Interactive Video 1
Interactive Video 2
Interactive Video 3
Interactive Video 4
Interactive Video 5
Kewar
Multiple Choice
Multiple Choice 2
personality quiz 1
Personality quiz 2
Sort Paragraphs
Summary
Timeline
True false
Virtual Tour 1
Virtual Tour 2

jankapunkt:
Wow, @sr258 you made an effort here. From my perspective this sounds good, especially with the scoping. We also use scss but I wonder if there are people who use just css or things like tailwind. Can they still use the theming then?

sr258:
Yes, the theme builder compiles it  to css.

PR #1974, opened 2021-12-12
feat(html-exporter)!: html exporter can display rights & licenses and embed code

sr258:
The HTML exporter can now display license information ("rights of use") and an embed code, if the user wishes. 

BREAKING CHANGES:
The template passed into the HTML exporter has a new signature. The integration object is now an IIntegration object instead of a string. You can use JSON.stringify(integration) to get the old string object.


JPSchellenberg (approved):
I tested it and it works just it should. Seems good to me and ready to merge.

PR #2100, opened 2022-02-25
perf(h5p-server): reduce memory and IO when uploading package

cabauman:
Reference a single file for identical file paths rather than one file for each. This improvement is limited to content types with copy/paste functionality.

So this came about when I was interacting with a Book Maker activity that has numerous images with 20 copies each (along with audio files for picked up and dropped). When I exported from WordPress and uploaded to Lumi, I noticed a given image/audio count jumps from one to the number of copies. On the other hand, creating a _new_ activity in Lumi with copy/pasted images references a single file. So it's only the upload feature that does this. Next, I tried the reverse to see how WordPress handles uploaded packages, and discovered that one image file remained as one file. So I figured it would be nice to bring that optimization here too.

I tested my changes manually, and it seems to work well. The following screenshot is the result of uploading a package with 8 copy/pasted images before and after the change:

Screenshot 2022-02-25 233528

If you guys welcome this change, I'll look into adding tests to go along with this PR. Off to bed for now!

_Thanks for all the hard work on this project._


otacke:
Just adding: BookMaker is a non-public content type that's re-purposing CoursePresentation: Removing plenty of things, adding some new, but basically CoursePresentation. Might help for reproducing the initial issue.

cabauman:
Thanks, Oliver. Can't keep track of them all 😅

sr258:
@cabauman Thank you for the PR! It looks like a very sensible addition and I'd be happy to merge it. If you could add a test that covers the new lines, it is ready to be merged.

packages/h5p-server/src/ContentStorer.ts
```diff
@@ -236,12 +236,18 @@ export default class ContentStorer {
                 )
             );
 
+        const filepathTonewFilenameMap = new Map<string, string>();
```
sr258: should probably be filepathTo**N**ewFilenameMap

cabauman: Good catch. I'll also go ahead and change the filepath part to file**P**ath for consistency with the rest of the code base.

cabauman:
Great. Let me know if there's anything you'd like me to tweak in the tests.

otacke:
@cabauman Thanks that you are donating your time!

cabauman:
@otacke It's a pleasure. I love the open source community.

sr258 (approved):

sr258:
Thank you for the contribution!

PR #2605, opened 2022-10-22
fix(package): remove duplicate scripts

borgiman:


sr258:
@borgiman Thanks for the PR!

PR #3191, opened 2023-09-19
2688 contentFilesUrlPlayerOverride does not load from H5PConfig

### LINKED ISSUE #2688: contentFilesUrlPlayerOverride does not load from H5PConfig
`contentFilesUrlPlayerOverride` does not load from the H5PConfig because it is missing in the `load()` and `save()-method:

[contentFilesUrlPlayerOverride
](https://github.com/Lumieducation/H5P-Nodejs-library/blob/release/packages/h5p-server/src/implementation/H5PConfig.ts#L117)

JPSchellenberg:
Hey,

the `contentFilesUrlPlayerOverride` option is not loaded from a config file. The option is missing in the `load()` and `save()` methods of the H5PConfig. This feature is needed to enable the content delivery directly from the object-storage:
https://docs.lumi.education/npm-packages/h5p-mongos3/mongo-s3-content-storage#increasing-scalability-by-getting-content-files-directly-from-s3

https://lumi-education.canny.io/bug-report/p/images-and-audio-clips-not-working

I added these and it works on the test-server, but I haven't found tests for H5PConfig. Are these missing or did I just not see them?

### DIFF
--- a/packages/h5p-server/src/implementation/H5PConfig.ts
+++ b/packages/h5p-server/src/implementation/H5PConfig.ts
@@ -116,6 +116,7 @@ export default class H5PConfig implements IH5PConfig {
      */
     public async load(): Promise<H5PConfig> {
         await this.loadSettingFromStorage('baseUrl');
+        await this.loadSettingFromStorage('contentFilesUrlPlayerOverride');
         await this.loadSettingFromStorage('contentHubEnabled');
         await this.loadSettingFromStorage('contentHubMetadataRefreshInterval');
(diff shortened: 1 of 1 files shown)


sr258 (approved):
Everything ok with the PR!

There are no tests. The config class is super simple, so I don't see reason to add tests for it.

sr258:
I'm not terribly convinced by how the H5PConfig class works anymore. It would probably be better to get the list of property names dynamically using decorators and reflection or something like that. That way it would be future proof.

PR #3195, opened 2023-09-20
Release next version

JPSchellenberg:


sr258 (dismissed):
The dependency updates shouldn't cause any problems and the only functional change is very minor. I haven't checked the diff, but the release should be fine!

JPSchellenberg:
I had to update the master branch due to merge conflicts. The release branch needs to be remerged into the master branch after a successful release. Maybe we switch back to releasing rom master?

sr258:
No, I don't think the the master branch should not be released immediately. We need some place to prepare releases when they should contain multiple merge requests and the Renovate bot is very active and this would lead to too many releases.

It would be better if the release process also updated the master branch and not just the release branch. The problem can also be solved by immediately pushing the release branch's head to the master, but this must be done manually and is easily forgotten.

PR #3320, opened 2023-11-01
feat: set `embedCode` and `resizeCode` when render player model

donnguyen:
Thanks for the great job.

This PR is to set `embedCode` and `resizeCode` when rendering player model.

### DIFF
--- a/packages/h5p-server/src/H5PPlayer.ts
+++ b/packages/h5p-server/src/H5PPlayer.ts
@@ -163,6 +163,8 @@ export default class H5PPlayer {
             showCopyButton?: boolean;
             showDownloadButton?: boolean;
             showEmbedButton?: boolean;
+            embedCode?: string;
+            resizeCode?: string;
             showFrame?: boolean;
             showH5PIcon?: boolean;
             showLicenseButton?: boolean;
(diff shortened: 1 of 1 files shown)


sr258:

Small, purely additive change — adds `embedCode`/`resizeCode` passthrough to `H5PPlayer.render()`/`generateIntegration()`. The `IIntegration.embedCode`/`resizeCode` fields already existed in `types.ts` (unused until now), so no type changes were needed, and no existing call site passes these options yet, so nothing downstream breaks.

**Findings:**

1. **`embedCode?.replace(':contentId', contentId)` only substitutes the first occurrence of the placeholder** (`H5PPlayer.ts:468`) — `String.prototype.replace` with a string pattern (not a `RegExp` with `/g`) is non-global. If an implementer configures `options.embedCode` with the placeholder appearing twice (e.g. a URL with `:contentId` in both the path and a query param), only the first is substituted, leaving a broken URL in the rendered embed code.

2. **Missing JSDoc for the two new `render()` options** (`H5PPlayer.ts:159`) — per CLAUDE.md ("JSDoc on all public methods and classes"), the other `options.*` params (contextId, asUserId, readOnlyState, etc.) are documented, but `options.embedCode`/`options.resizeCode` aren't, leaving consumers without guidance on the `:contentId` placeholder convention.

Both are low severity; #1 is worth a fix if `:contentId` could plausibly appear more than once in a template.

sr258:
Thanks for this @donnguyen! I picked this up, rebased it onto current master, and found/fixed a small bug during review: the `:contentId` placeholder substitution used `String.prototype.replace`, which only replaces the first occurrence — templates referencing `:contentId` more than once (e.g. in both a URL and an attribute) would leave the literal placeholder in the output. Switched to `replaceAll`, added JSDoc for the new `embedCode`/`resizeCode` render options, and added a regression test.

Continued in #4613 — closing this one in favor of that.

PR #3703, opened 2024-07-10
[H5P Server] Add Configuration Overrides for coreApiVersion and h5pVersion

dominikbruechner:
### Summary
This PR introduces the ability to override the H5P Server's `coreApiVersion` and `h5pVersion` configuration values from storage, which should be currently set to 1.26 and 1.25, respectively.

### Motivation
By allowing these overrides, we enable updating the core (as outlined in [Lumi's core updates documentation](https://docs.lumi.education/development/core-updates)) without requiring users to manually override the config, but load the values from storage.

### Changes

- Added configuration options for coreApiVersion and h5pVersion.

Cheers! :)


packages/h5p-server/src/implementation/H5PConfig.ts
```diff
@@ -122,12 +122,14 @@ export default class H5PConfig implements IH5PConfig {
         await this.loadSettingFromStorage('contentTypeCacheRefreshInterval');
         await this.loadSettingFromStorage('contentUserStateSaveInterval');
         await this.loadSettingFromStorage('contentWhitelist');
+        await this.loadSettingFromStorage('coreApiVersion');
         await this.loadSettingFromStorage('customization');
```
sr258, line 126: `coreApiVersion` and `h5pVersion` are now loaded from/saved to storage here, but the JSDoc for both fields in `packages/h5p-server/src/types.ts` (lines ~1671-1676 and ~1745-1749) still states: "Not user-configurable and should not be changed by custom implementations." That comment should be updated to reflect the new override capability — otherwise implementers reading the type definitions will believe these fields are immutable, while in practice a stale/incorrect stored value (e.g. a `coreApiVersion` that doesn't match the core JS/CSS actually deployed at `coreUrl`) will silently change what `PackageValidator.checkCoreVersion` accepts, potentially allowing libraries that require APIs not present in the physically deployed core files.

sr258:
Note: the review comment above was generated by Claude Code (automated review), not a manual read-through by me.

sr258:
Thanks for the PR, @dominikbruechner — and sorry for the delayed review.

After looking closely at this, we've decided not to merge it. The core issue is that `coreApiVersion` and `h5pVersion` aren't really independent settings you can safely repoint at another value — they're metadata that's supposed to describe whatever core files are actually bundled with the installed `h5p-server` package:

1. **Overriding the version doesn't change which core files get served.** The list of core JS/CSS files served to the player/editor comes from a static asset manifest baked into the package (`playerAssetList` in `H5PPlayer`/`H5PEditor`), not from `config.coreApiVersion`/`h5pVersion`. So even with this override in place, pointing the version at, say, a newer core release wouldn't actually make the server serve that newer core — it would just make the reported version lie about what's really there. That defeats the likely purpose of the feature.

2. **`coreApiVersion` gates package validation.** `PackageValidator.checkCoreVersion()` uses it to decide whether an uploaded `.h5p` package is compatible with the installed core. If this value is persisted in storage and drifts from the actual bundled core (e.g. left over from before an `h5p-server` upgrade), it can silently accept packages the real core can't handle, or reject ones it could.

3. **`h5pVersion` is used as a cache-buster.** It's appended to core/editor asset URLs specifically so that upgrades bust old caches. Persisting it means a stored value can outlive an upgrade, so browsers/CDNs keep serving stale, pre-upgrade core files with no error surfaced anywhere.

4. Both fields are already documented as "Not user-configurable and should not be changed by custom implementations" — and implementers who genuinely need a different value for testing purposes already have that option today via a custom `IH5PConfig` implementation, without the drift risk that comes from wiring them into the generic storage-backed load/save cycle used for the actual user-configurable settings.

If there's a real use case behind this (e.g. testing against a specific core version, or supporting a pinned/vendored core build), we'd be glad to discuss an approach that makes the asset manifest and the reported version consistent with each other, rather than letting them diverge. Happy to reopen the conversation if you want to go that route.

Closing this PR for now — thanks again for the contribution!

PR #4102, opened 2025-05-22
fix(h5p-server): fix the package exporter so it also exports addOns

biwano:
When exporting a content, the current implementation do not include the library storage's addOns.
In our case, this prevented content that required mathJax to be exported properly.

This PR (probably) fixes this.

cc @sr258 
cc @JPSchellenberg

### DIFF
--- a/packages/h5p-server/src/PackageExporter.ts
+++ b/packages/h5p-server/src/PackageExporter.ts
@@ -11,7 +11,8 @@ import {
     IContentMetadata,
     IUser,
     ContentPermission,
-    IPermissionSystem
+    IPermissionSystem,
+    ILibraryMetadata
 } from './types';
 import { ContentFileScanner } from './ContentFileScanner';
 import Logger from './helpers/Logger';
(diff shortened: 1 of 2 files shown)


sr258:

Reviewed `packages/h5p-server/src/PackageExporter.ts`. No tests cover the new addon-export path.

**Findings:**

1. **`LibraryName.fromUberName(addOn.machineName)` is called with a bare machine name (e.g. `"H5P.MathDisplay"`), not an uber-name with a version suffix — parsing fails.** (`PackageExporter.ts:190`)
   Whenever any addon is installed (`listAddons()` returns non-empty), `addOn.machineName` doesn't match the uber-name regex `/^([\w.]+)-(\d+)\.(\d+)$/i`, so `LibraryName.fromUberName` throws `H5pError('malformed-ubername', ..., 400)`. This causes `createPackage()` to reject for *any* content export as soon as an addon is present — exactly the scenario (mathJax addon) this PR is meant to fix.

2. **New private method `addAddonsFiles` is dead code.** (`PackageExporter.ts:210`)
   It duplicates the addon-file-adding logic already inlined in `addLibraryFiles`, but there are no call sites — it's defined but never invoked, and its `ILibraryMetadata[]`-typed dependency list (vs. `addLibraryFiles`'s `ILibraryName[]`) will silently drift out of sync since nothing exercises it.

Issue #4197, opened 2025-07-22
HTML Export no longer works with latest H5P.Video


JPSchellenberg:
The HTML Export does not work with the latest version of H5P.Video:

H5P.Video has a script called echo360.js which uses the delete statement on an expression:

https://github.com/h5p/h5p-video/blob/master/scripts/echo360.js#L229

uglifyjs can therefore not parse the script file as it uses strict mode and throws the following error: 
JS_Parse_Error: Calling delete on expression not allowed in strict mode

As far as I have seen it is not possible to turn strict mode off. Therefore we have limited options:

1) Remove uglifyjs and just join all scripts in the array with array.prototpe.join. I tested this with a file and it seems to work. The code is no longer minified and optimised.

2) Use something different than uglifyjs. I would have to look up alternatives.

sr258:
Does simply concatenating all scripts with join really solve the issue? I would assume that the 'use strict' statement of another file makes the whole bundle become a file in strict mode and that this is why the delete statement fails. If this is the case, using another minification library might not solve the problem. I think that having a minifier is pretty important, as it really reduces file sizes. It shouldn't be too hard to replace uglifyjs.

Another solution might be to remove all 'use strict' from the bundle. There is some risk, that this will break something, though. 

I've had an issue with strict mode when updating uglifyjs some time ago: https://github.com/lumieducation/h5p-nodejs-library/commit/b0d7e4d13fd27b5cd627def6803d7d412821c001 Maybe it is somehow possible to encapsulate the scripts into module and limit the effects of strict mode to a module?

JPSchellenberg:
I think removing all 'use strict' from the bundle might have side effects down the road and would not do it. I would have to check what it does to the file size, but my gut feeling is that it is quite helpful to minimise and also parse the scripts to see if they work. 

Another solution would be to contact the maintainer of H5P.Video in hope they fix their script to comply in strict mode. I think that would be the most elegant solution, but I have no idea how long it takes them to fix that. Until then the HTML Export with videos does not work which seems to be quite a big problem for a lot of people.

sr258:
@JPSchellenberg How have you proceeded here?

GitGianluc:
[...]

PR #4377, opened 2026-01-11
fix: patch echo360.js on the fly to avoid error in strict mode

mindcrunch4u:
Root Cause:
- `echo360.js` uses bad syntax, which causes `uglify-js` to throw error in `strict mode`.

Changes:
- Add a piece of logic to patch `echo360.js` during export.

cc Original Author: @sr258 

---

Regarding [this issue](https://github.com/Lumieducation/H5P-Nodejs-library/issues/4197):
> Another solution would be to contact the maintainer of H5P.Video in hope they fix their script to comply in strict mode. I think that would be the most elegant solution

I've created a PR here: [link](https://github.com/h5p/h5p-video/pull/158), let's hope [otacke](https://github.com/otacke) fixes this issue soon.

Meanwhile I'd like this propose a more flexible fix (this PR):
- Since it's impractical to make h5p authors to write perfect code, we could instead fix faulty code on-the-fly before passing them through `uglify-js`.
- With this kind of design, we can respond to our user as soon as possible by delivering patches (instead of waiting for h5p).
- Of course, we could file an issue to h5p's library at the same time, and if the issue gets fixed, then we could remove our patches in subsequent releases.


mindcrunch4u:
Adding @JPSchellenberg for review since this also affects Lumi users.

otacke:
@mindcrunch4u Just noticed my name here by chance: I agree that this should be fixed in H5P.Video, but I am not the maintainer of that repository and cannot push anything. You will need to reach out to H5P Group.

sr258:
@mindcrunch4u Thank you for your PR. I've picked this up and generalized the patch mechanism. Unfortunately, I can't reproduce the issue to check whether the issue is actually solved by the PR. Can you give me a reproducible instruction how to cause the bug?

mindcrunch4u:
@sr258 to reproduce, I simply added a video to a basic H5P project, then export as SCORM. See [this issue](https://github.com/Lumieducation/Lumi/issues/2692#issuecomment-3736771124).

Your `LibraryPatcher` looks good! If it works, then Lumi should function like this: see video in [this comment](https://github.com/Lumieducation/Lumi/issues/2692#issuecomment-3734644628).

To verify using Lumi, this is my write-up: [link](https://icandothese.com/docs/tech/development_front/patch_an_npm_package/#using-the-patch).

Basically:

1. Upload the patched version of `H5P-Nodejs-library` to npm
2. Get Lumi source code
    ```
    nvm install 24
    git clone https://github.com/Lumieducation/Lumi.git
    ```
3. Then edit `package.json`, change `@lumieducation/h5p-html-exporter` to your new npm repo (containing the updated `H5P-Nodejs-library`)
4. `npm install` and `npm run build` then `npm start`.

mindcrunch4u:
@sr258 I've done the verification. My steps:
- upload your (@sr258 ) version of `h5p-html-exporter` to npm: [link](https://www.npmjs.com/package/@mindcrunch4u/h5p-html-exporter)
- update Lumi's `package.json` to point to the updated `h5p-html-exporter`

Here's the comparison.

Without this patch (containing your commits), Lumi hangs when exporting the project:


https://github.com/user-attachments/assets/6d444472-4388-4464-838e-0a801358e4f5


With the patch:

https://github.com/user-attachments/assets/67c545d3-d9e7-411a-828e-9bbc91ad2fb2

Note: because of changes to `HtmlExporter` (I saw you added `options?: IHtmlExporterOptions`), I did a bit of hack [here](https://github.com/Lumieducation/Lumi/blob/master/src/ops/export-h5p.ts#L216) to make the code run (removed `options` when calling `HtmlExporter` from Lumi).

sr258:

Note: the diff also includes a large `.circleci/config.yml` hunk that's stale drift unrelated to this PR (the repo has no `.circleci/` directory at all in the current tree — CI runs via `.github/workflows/ci.yml` per CLAUDE.md). Excluded that from review as out of scope; worth rebasing to drop it.

**Findings:**

1. **`HtmlExporter` constructor's new `options?: IHtmlExporterOptions` param silently breaks existing callers instead of accepting both forms** (`packages/h5p-html-exporter/src/HtmlExporter.ts:279`) — the pre-PR signature took positional `template?`/`translationFunction?` params. A caller still doing `new HtmlExporter(ls, cs, config, coreDir, editorDir, myTemplateFn)` still type-checks, because `IHtmlExporterOptions` has only optional properties and a function is structurally assignable to it. At runtime `options.template` is `undefined`, so the custom template is silently dropped and the default is used — no compile error, no runtime error, just wrong output. This PR is tagged `fix:` with no `BREAKING CHANGE:` footer, so semantic-release would only bump patch/minor, shipping this break to consumers on caret ranges.

2. **`applyPatches` ignores `replaceAll: false` when the patch's `search` RegExp already carries the `g` flag** (`packages/h5p-html-exporter/src/helpers/LibraryPatcher.ts:516`) — contradicts the documented contract on `ILibraryFilePatch.replaceAll` in `types.ts` ("If false or omitted, only the first occurrence is replaced. @default false"). The code only strips/adds the `g` flag when `replaceAll` is `true`; for `replaceAll: false` it preserves whatever flag the input regex had, so a patch author writing `{ search: /foo/g, replace: 'bar' }` without `replaceAll: true` gets every occurrence replaced, not just the first. This is actually asserted (not caught) by this PR's own new test `'replaces all matches when regex already has g flag'` in `LibraryPatcher.test.ts` — worth double-checking whether that test encodes the intended behavior or the bug.

PR #4386, opened 2026-01-28
chore: update content type hub endpoint

otacke:
When merged in, will update references to the H5P content type Hub server.

H5P Group updated their endpoints: https://github.com/h5p/h5p-php-library/commit/69c25bf67671e25ceafc26358b7dfc9c8be6aeed


sr258:
@otacke Thanks for making us aware of this change. It looks like this is not just an endpoint change but the new API behaves differently from the old one:

- The new endpoint return status code 422 when it is called with a local id longer than 15 characters.
- The h5p packages that you download from the endpoints only contain library directories, but no h5p.json and dummy content like before.

I've changed the library so that it only uses up to 15 characters for the local id, but not the other API change. Before I do that I'd like to understand this new API a bit better: Do you know anything about why it's there?

otacke:
Sorry!
Neither did I check the contents of the packages you get nor did I know any details.

otacke:
@sr258 
Sorry again, should have checked before. But the only thing that was changed in H5P core was the subdomain, nothing else (the request behavior in particular), and I expected the node.js port already met the requirements.

- Sending a `POST` to `hub-api.h5p.org/v1/content-types/` while providing at least a `uuid` in the body still serves the hub cache data like before.
- Sending a `POST` to `hub-api.h5p.org/v1/content-types/$MACHINE_NAME` while providing a `uuid` in the body now in fact fails for fetching the respective content type

When using `api.h5p.org/v1/content-types/$MACHINE_NAME` to fetch a library archive, no additional data was necessary in the body.

Now, when sending a `POST` request to `hub-api.h5p.org/v1/content-types/$MACHINE_NAME`, even if I provide a `local_id` and the other registration data keys/values, I do not even get past a 403. What may I be missing?

Either way: The `local_id` should be a CRC32 hash, so would be represented by an 8-character hexadecimal string. Accepting up to 15 characters seems to be generous then. An SHA-256 hash as generated by `machineIdSync` with 64 hexadecimal characters is too long.

Regarding the package contents: H5P library files are supposed to be used for installing libraries, so they do not require `h5p.json` or `content.json` or dummy content (likewise, content files do not strictly require the library folders, but only `h5p.json` (to define what the content needs) and `content.json` (as the parameters), and media files should be provided as well. If library folders are not part of the H5P content file, then the respective libraries will have to be installed already, or the content cannot be uploaded.

Once more: I am terribly sorry for not having checked the new endpoints before.

sr258:
No worries! I've looked closer at the change and the H5P server itself handles the content-free H5P files fine. The issue is that the library uses these files for unit tests. I'll think of a solution for that.

otacke:
@sr258 I see.

The only thing I still don't get is why I hit a 403 why trying to fetch a content type library (all registration data set). Similarly, even when calling the SITES endpoint with POST, the response claims I used get: 405, `The GET method is not supported for route v1/sites.`

Have you encountered anything similar? I'd assume the server is having a different problem with my request but fails to tell me the truth.

sr258:
@otacke For me the only issue was the too long localId, but I didn't have any other issues.

otacke:
Sorry for misusing this comment. Here's what I found (including a solution).

# Trying to get a UUID
## Old endpoint
```
curl --location --request POST 'api.h5p.org/v1/sites'
```
yields a UUID, e.g.
```
{
    "uuid": "019c7bec-b95b-70a4-a594-cda2ed91db7d"
}
```
Would work with `GET` as well.

## New endpoint
```
curl --location --request POST 'hub-api.h5p.org/v1/sites' \
--form 'platform_name="WordPress"' \
--form 'platform_version="6.9"' \
--form 'h5p_version="1.16.2"' \
--form 'disabled="0"' \
--form 'local_id="1a2b3c4d"' \
--form 'type="local"' \
--form 'core_api_version="1.27"'
```
also yields a UUID, e.g.
```
{
    "uuid": "019c7bec-b95b-70a4-a594-cda2ed91db7d"
}
```
So, with the new endpoint, the additional data seems to be mandatory. If you don't then you receive the misleading error message
```
{
    "message": "The GET method is not supported for route v1/sites. Supported methods: POST."
}
```
And here, despite you `GET`ting data, `GET` does not work.

# Trying to get overview of content types served
## Old endpoint
```
curl --location --request POST 'api.h5p.org/v1/content-types' --form 'uuid="40380b7c-74f6-45bc-a4ff-b648854343b1"'
```
yields the overview

## New endpoint
```
curl --location --request POST 'hub-api.h5p.org/v1/content-types' --form 'uuid="40380b7c-74f6-45bc-a4ff-b648854343b1"'
```
also yields the overview, even without supplying the additional data. So the new server is  handling things a little inconsistently. Using `GET` does not work, even though you are `GET`ting data only.

# Trying to get GameMap
## Old endpoint
```
curl --location --request POST 'api.h5p.org/v1/content-types/H5P.GameMap' \
--form 'uuid="40380b7c-74f6-45bc-a4ff-b648854343b1"'
```
yields the `.h5p` file

## New endpoint
```
curl --location --request POST 'hub-api.h5p.org/v1/content-types/H5P.GameMap' \
--form 'uuid=019c7bec-b95b-70a4-a594-cda2ed91db7d"' \
--form 'platform_name="WordPress"' \
--form 'platform_version="6.9"' \
--form 'h5p_version="1.16.2"' \
--form 'disabled="0"' \
--form 'local_id="1a2b3c4d"' \
--form 'type="local"' \
--form 'core_api_version="1.27"'
```
yields
```
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
<HTML>

<HEAD>
	<META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=iso-8859-1">
	<TITLE>ERROR: The request could not be satisfied</TITLE>
</HEAD>

<BODY>
	<H1>403 ERROR</H1>
	<H2>The request could not be satisfied.</H2>
	<HR noshade size="1px">
	Bad request.
	We can't connect to the server for this app or website at this time. There might be too much traffic or a
	configuration error. Try again later, or contact the app or website owner.
	<BR clear="all">
If you provide content to customers through CloudFront, you can find steps to troubleshoot and help prevent this error by reviewing the CloudFront documentation.
	<BR clear="all">
	<HR noshade size="1px">
	<PRE>
Generated by cloudfront (CloudFront)
Request ID: TmkYJ4kc2cIq5cYtocLBVbea-WIm_Stk4mzvghlL8UhjkFO3iU5txg==
</PRE>
	<ADDRESS>
	</ADDRESS>
</BODY>

</HTML>
```

Here, for whatever reason, you are expected to use `GET`, not `POST`, and supply a `UUID` and the additional data.

```
curl --location --request GET 'https://hub-api.h5p.org/v1/content-types/H5P.GameMap?uuid=019c7bec-b95b-70a4-a594-cda2ed91db7d&platform_name=WordPress&platform_version=6.9&h5p_version=1.16.2&disabled=0&local_id=1a2b3c4d&type=local&core_api_version=1.27'
```
will in fact get you what you want (need to use `--output` to actually get the file, of course).

PR #4600, opened 2026-09-06
Feat/buffer sanitizer and malware scanners

sr258:
Based on #4362 (own branch because we adopted the PR).

## What this does

Until now, file sanitization (`IFileSanitizer`) and malware scanning
(`IFileMalwareScanner`) only ran for uploads that were backed by a temporary
file on disk. Integrators whose upload middleware keeps files in memory (e.g.
`express-fileupload` without `useTempFiles`) silently got **no** SVG
sanitization and **no** virus scanning — the docs stated this limitation, but
nothing in the code signalled it at runtime.

This PR makes both plugin types work for in-memory buffers as well, so
`H5PEditor.saveContentFile` behaves the same regardless of how the integrator
receives uploads.

## The interfaces

`H5PFile` and `H5PFileBuffer` are now exported from `h5p-server` and used
everywhere an uploaded file is passed around (previously the same anonymous
`{ data?, mimetype, name, size, tempFilePath? }` shape was inlined in
`H5PAjaxEndpoint`, `h5p-express`'s `IActionRequest`, and elsewhere).

Both plugin interfaces gain an **optional** buffer method:

```ts
interface IFileMalwareScanner {
    scan(file: string): Promise<{ result: MalwareScanResult; viruses?: string }>;
    scanBuffer?(file: H5PFileBuffer): Promise<{ result: MalwareScanResult; viruses?: string }>;
}

interface IFileSanitizer {
    sanitize(file: string, originalFilename?: string): Promise<FileSanitizerResult>;
    sanitizeBuffer?(file: H5PFileBuffer): Promise<FileSanitizerResult>;
}
```

Both additions are backwards compatible: **existing implementations keep
working unchanged.** If a plugin doesn't implement the buffer method,
`H5PEditor` writes the buffer to a temporary file, calls the path-based
method, and reads the result back. When several path-based scanners run on the
same buffer, that temporary file is written **once** and shared between them.

`sanitize` additionally takes an optional `originalFilename`. Upload
middlewares typically generate extensionless temp paths
(`tmp-5000-1567887897`), so a sanitizer that decides what to do based on the
file name needs the *original* name, not the path it was handed. It defaults
to `file`, so existing callers and implementations are unaffected.

## Buffer vs. temp file precedence

Every code path that has to choose between `file.data` and `file.tempFilePath`
— validation, sanitization, malware scanning and persistence — now goes
through one shared `hasBufferData()` helper, so they can no longer disagree
about what the actual file content is:

- `data` wins whenever it carries content;
- a *defined but empty* `data` alongside a real `tempFilePath` (an
  `express-fileupload` `useTempFiles` quirk — it always sets
  `data = Buffer.concat([])`) defers to the file on disk;
- a defined empty `data` with no `tempFilePath` (a genuine 0-byte buffer-only
  upload) is authoritative, since there is nothing to fall back to.

Before this, `sanitizeFile`/`scanForMalware` preferred `tempFilePath` while
validation and persistence preferred `data`. For middleware that populates
both, the sanitizer rewrote the temp file while the *unsanitized* buffer was
what actually got stored — an XSS-carrying SVG would have been persisted
unsanitized.

## Content validation

`contentFileValidation` now handles buffers (`validateBufferContent`) in
addition to paths, and the claimed extension is always derived from the
**original filename**, never from a middleware-generated path. Deriving it
from an extensionless temp path made `extensionMatchesDetected('', [...])`
fail every tier, so any file whose magic bytes *were* recognized got rejected
with `upload-validation-error` — i.e. in a `useTempFiles: true` setup (what
both example apps use) no image, audio or video could be uploaded at all.

A file whose original filename has no extension is now rejected outright in
both the buffer and the temp-file path. This is belt-and-braces —
`H5PEditor.saveContentFile`'s whitelist check and
`PackageValidator.isAllowedFileExtension` already reject those upstream — but
it makes the validator fail closed instead of depending on its callers. Side
effect: for an extensionless editor upload the surfaced error is now
`upload-validation-error` rather than `not-in-whitelist`, because validation
runs before the whitelist check.

## Package changes

**`h5p-svg-sanitizer`** implements `sanitizeBuffer`. Its SVG detection now
uses `originalFilename` (falling back to the path) and is case-insensitive, so
`image.SVG` is sanitized too. Previously it decided SVG-ness from the path it
was given, which for a real editor upload is the extensionless temp path — so
in exactly the setup the docs recommend for malware scanning, every SVG upload
returned `Ignored` and was never sanitized.

**`h5p-clamav-scanner`** implements `scanBuffer`, streaming the buffer
directly to the clamd daemon when one is configured, and otherwise falling
back to a temporary file. Whether a daemon is available is read back from
clamscan's *resolved* settings and additionally requires a resolved
`socket`/`port`/`host`: clamscan's `scanner` defaults to `'clamdscan'` even
with no connection configured (it then shells out to the local binary), and
`scanStream` rejects with *"A unix socket or port (+ optional host) is
required!"* in that case. `scanBuffer` caught that rejection and returned
`NotScanned`, and `saveContentFile` only rejects on `MalwareFound` — so an
infected buffer upload was accepted with no scan and no error.

## Also included

- Docs (`docs/advanced/security.md`, `h5p-clamav-scanner/README.md`) no longer
  claim that sanitization/scanning requires temporary files.
- Unrelated fix carried over from #4362: `LibraryManager` reported
  `installLibraryLockTimeout` in the
  `server:install-library-lock-max-time-exceeded` error where it meant
  `installLibraryLockMaxOccupationTime`.
- `h5p-rest-example-server` route params are `.toString()`-ed for the
  Express 5 typings.

## Testing

+2318/-374 across 21 files, roughly half of it tests. New coverage for the
buffer paths in `contentFileValidation`, `H5PEditor.saving`, `SvgSanitizer`
and `ClamAVScanner`, including: extensionless filenames (rejected),
extensionless temp paths with a proper `file.name` (accepted), genuine 0-byte
buffer-only uploads, SVG sanitization driven by `originalFilename`, and
regressions proving `file.data` rather than a stale `tempFilePath` is what
gets sanitized and scanned when both are present.

All CI checks pass, including `clamav-tests`, which runs the scanner suite
against a real `clamav/clamav` daemon service.

### DIFF
--- a/docs/advanced/security.md
+++ b/docs/advanced/security.md
@@ -87,7 +87,7 @@ interface IFileSanitizer {
 
     /** Sanitizes files. The original file is expected to be replaced by the
      * sanitized file, so there is no new path to the sanitized file.*/
-    sanitize(file: string): Promise<FileSanitizerResult>;
+    sanitize(file: File): Promise<FileSanitizerResult>;
 }
 
 enum FileSanitizerResult {
@@ -97,9 +97,10 @@ enum FileSanitizerResult {
 }
 ```
 
-Note: Sanitization only works if you pass uploaded content files to {@link
-@lumieducation/h5p-server!H5PEditor.saveContentFile} as temporary files, not as
-in-memory streams!
+ Note: Sanitization works for uploaded content files that you pass to {@link
+ @lumieducation/h5p-server!H5PEditor.saveContentFile} using the {@link
+ @lumieducation/h5p-server!File} shape, regardless of whether they are backed by
+ temporary files on disk or in-memory buffers.
 
 ### Existing file sanitizers
 
@@ -175,7 +176,7 @@ interface IFileMalwareScanner {
 
     /** Scans a file for malware and returns whether it contains malware. */
     scan(
-        file: string
+        file: File
     ): Promise<{ result: MalwareScanResult; viruses?: string }>;
 }
 
--- a/docs/advanced/security.md
+++ b/docs/advanced/security.md
@@ -85,9 +85,20 @@ interface IFileSanitizer {
     /** The name of the scanner, e.g. SVG Sanitizer. Used in debug output */
     readonly name: string;
 
-    /** Sanitizes files. The original file is expected to be replaced by the
-     * sanitized file, so there is no new path to the sanitized file.*/
-    sanitize(file: File): Promise<FileSanitizerResult>;
+    /** Sanitizes the file at the given path. The original file is expected
+     * to be replaced by the sanitized file, so there is no new path to the
+     * sanitized file. */
+    sanitize(file: string): Promise<FileSanitizerResult>;
+
+    /**
+     * Optional: sanitizes an in-memory buffer without requiring it to be
+     * written to disk first. Implementations are expected to replace
+     * `file.data` with the sanitized buffer. Implementations that don't
+     * support buffer sanitization can omit this method; callers fall back
+     * to writing the buffer to a temporary file and calling `sanitize`
+     * instead.
+     */
+    sanitizeBuffer?(file: H5PFileBuffer): Promise<FileSanitizerResult>;
 }
 
 enum FileSanitizerResult {
@@ -97,10 +108,13 @@ enum FileSanitizerResult {
 }
 ```
 
- Note: Sanitization works for uploaded content files that you pass to {@link
- @lumieducation/h5p-server!H5PEditor.saveContentFile} using the {@link
- @lumieducation/h5p-server!File} shape, regardless of whether they are backed by
- temporary files on disk or in-memory buffers.
+Note: Sanitization works transparently for uploaded content files that you
+pass to {@link @lumieducation/h5p-server!H5PEditor.saveContentFile} using the
+{@link @lumieducation/h5p-server!H5PFile} shape, regardless of whether they
+are backed by temporary files on disk (`tempFilePath`) or in-memory buffers
+(`data`). If a sanitizer doesn't implement `sanitizeBuffer`, buffer-based
+uploads are automatically written to a temporary file before being passed to
+`sanitize`, so all sanitizers support both upload modes without extra work.
 
 ### Existing file sanitizers
 
@@ -174,9 +188,20 @@ interface IFileMalwareScanner {
     /** The name of the scanner, e.g. ClamAV */
     readonly name: string;
 
-    /** Scans a file for malware and returns whether it contains malware. */
+    /** Scans a file at the given path for malware and returns whether it
+     * contains malware. */
     scan(
-        file: File
+        file: string
+    ): Promise<{ result: MalwareScanResult; viruses?: string }>;
+
+    /**
+     * Optional: scans an in-memory buffer for malware without requiring it
+     * to be written to disk first. Implementations that don't support
+     * buffer scanning can omit this method; callers fall back to writing
+     * the buffer to a temporary file and calling `scan` instead.
+     */
+    scanBuffer?(
+        file: H5PFileBuffer
     ): Promise<{ result: MalwareScanResult; viruses?: string }>;
 }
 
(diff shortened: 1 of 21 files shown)


sr258:
## Code review follow-up (3 commits pushed)

A review of this branch turned up five issues, one of them a blocker. All five are fixed in `338e3a6b`, `029fcc28`, `2d51975c`.

### 1. Extension was derived from the temp path, breaking every temp-file upload (blocker)

`validateFileContent` took the claimed extension from the path it was handed, and `validateContent` passed `file.tempFilePath`. express-fileupload's temp files are extensionless (`tmp-5000-1567887897`), so `ext` was `''` and `extensionMatchesDetected('', ['png'])` failed all three tiers — no direct match, `mimeLookup('')` is falsy, no residual group. Any file whose magic bytes *were* recognized got rejected with `upload-validation-error`, i.e. no image, audio or video could be uploaded at all in the `useTempFiles: true` setup used by both `h5p-examples` and `h5p-rest-example-server`. The existing tests passed only because every fixture carries an extension.

The extension is now always derived from the **original filename**, never from a path that middleware generated. `validateFileContent` takes an optional `originalFilename` (defaulting to the path, so the `ContentStorer` callers that pass real in-package paths are unaffected), and `validateContent` fills it in from `file.name`.

A file whose original filename has **no extension at all is now rejected outright**, in both the buffer and the temp-file path. This is belt-and-braces — `H5PEditor.saveContentFile`'s whitelist check and `PackageValidator.isAllowedFileExtension` already reject extensionless files upstream — but it makes `contentFileValidation` fail closed instead of depending on its callers. One side effect worth noting: for an extensionless editor upload the surfaced error is now `upload-validation-error` rather than `not-in-whitelist`, because validation runs before the whitelist check.

### 2. Buffer/temp-file precedence disagreed between code paths

`sanitizeFile` and `scanForMalware` checked `tempFilePath` first, while `validateContent` and the persistence step in `saveContentFile` preferred `data`. For any middleware populating both, the sanitizer rewrote the temp file while the original unsanitized buffer was what actually got stored — an XSS-carrying SVG would have been persisted unsanitized.

All four paths now share one `hasBufferData()` helper. `data` wins when it carries content, or when it is defined with no `tempFilePath` at all; a defined-but-empty buffer alongside a real `tempFilePath` (express-fileupload's `useTempFiles` quirk) correctly defers to disk. In `scanAllForMalware` the shared-temp-file optimization strips `data` from the object handed to path-based scanners, otherwise the new precedence would have had each scanner write its own copy of the buffer again.

### 3. SVGs uploaded as temp files were never sanitized

`SvgSanitizer.sanitize` decided SVG-ness from `basename(file)`, and `H5PEditor` passed the extensionless temp path — so every real SVG upload returned `Ignored` in exactly the setup the docs recommend for malware scanning. (Pre-existing; the old `file.endsWith('.svg')` had the same flaw.) `IFileSanitizer.sanitize` now takes an optional `originalFilename` that `H5PEditor` fills from `file.name`; it defaults to the path, so existing implementations keep working.

### 4. ClamAV buffers could go silently unscanned

`clamdServiceEnabled` was derived from `scanner === 'clamdscan'`, but clamscan resolves that name even with no socket/host/port configured — it then shells out to the local binary, and `scanStream` rejects with *"A unix socket or port (+ optional host) is required!"*. `scanBuffer` caught the rejection and returned `NotScanned`, and `saveContentFile` only rejects on `MalwareFound`, so an infected buffer upload was accepted with no scan and no error. The flag now additionally requires a resolved `socket`, `port` or `host`, so buffers fall back to the temp-file scan.

### 5. Empty-buffer uploads were inconsistent

`file.data?.length > 0` sent a genuine 0-byte buffer-only upload into the error branch, while `validateFileContent` returns early on an empty file and `validateBufferContent` no-ops. Now consistent via `hasBufferData()`.

### Tests

Added coverage for: a valid PNG with an extensionless name (rejected), a temp-file upload with an extensionless path but a proper `file.name` (accepted), a temp-file upload whose `file.name` has no extension (rejected), a genuine 0-byte buffer-only upload (accepted), `SvgSanitizer` with an extensionless path plus `.svg` original filename (sanitized) and the inverse (ignored), and two regressions proving `file.data` rather than a stale `tempFilePath` is what gets sanitized and scanned when both are present.

Full suite green: 38 files, 356 tests. `build`, `lint` and `format:check` all pass.

**One gap:** no new test for the ClamAV fix — there is no `clamscan`/`clamdscan` binary available in this environment and `h5p-clamav-scanner` is not in the root vitest include list, so that package's suite could not be run here. The existing *"uses temp file scanning when no daemon config is provided"* test covers the scenario, but the fix itself is unverified by execution and should be exercised against a real clamd before merge.

bergatco:
Dear @sr258 ,

many, many thanks for taking this up! 🙇‍♂️ 
I think, it is a quite nice extension to your project and hope a lot of people will find it useful.

Regards,
Constantin Bergatt

PR #4614, opened 2026-09-08
test(h5p-e2e): scaffold Playwright E2E suite (sessions 1-10, plan complete)

sr258:
## Summary
Implements all 10 sessions of `E2E_AUTOMATION_PLAN.md` (a plan for replacing most of `test-plan.md` with an automated Playwright E2E suite driving the real `h5p-examples`/`h5p-rest-example-*` apps). This PR now completes the entire plan.

- **Session 1** — scaffolds a new private `packages/h5p-e2e` workspace, Playwright config (Chromium, single worker, `webServer` starting `h5p-examples`), root `test:e2e`/`test:e2e:ui` scripts, and a smoke test asserting the start page renders. Also fixes a pre-existing bug where `keyv` wasn't hoisted to the root `node_modules`, which crashed `packages/h5p-examples` at startup (`@keyv/redis` resolves it as a peer dependency from the root).
- **Session 2** — adds page objects (`StartPage`, `EditorPage`, `PlayerPage`, `LibraryAdminPanel`, `ContentTypeCachePanel`) with selectors verified empirically against a running instance, documented in `packages/h5p-e2e/SELECTORS.md`, plus a spec that creates, saves and plays a minimal H5P.Blanks content end to end. Also corrects a wrong assumption in the session 4 plan (the content type cache panel has no download button).
- **Session 3** — adds `test/fixtures/{reset,seed,index}.ts`: a filesystem state resetter behind an `IStateResetter` interface (so a session 7 Mongo/S3 implementation only needs to add a class), a `seedLibraries()` helper that installs content types from local `.h5p` fixtures via the ajax `library-upload` endpoint, and a `test` extended from `@playwright/test` that fails any spec on an unexpected browser console error or uncaught page exception (an allowlist exists for deliberately tolerated messages). Corrects a wrong assumption in the plan: Playwright starts `webServer` *before* running `globalSetup`, not after, so the reset can't live in a `globalSetup` hook as originally planned — it's chained in front of the actual start command in `webServer.command` instead. The session-2 Blanks spec now uses `seedLibraries()` instead of its own inline setup.
- **Session 4** — adds `test/specs/library-management.spec.ts` covering the "Library management" section of `test-plan.md`: an empty admin panel right after a storage reset, installing content types via the seeding fixture with version-number assertions, deleting a content type through the GUI, updating the content-type cache via "Update now" (`@network`), and installing the H5P.MathDisplay addon (`@network`) with `isAddon` asserted via `GET /h5p/libraries`. Two more plan corrections are documented inline: the admin panel only shows a delete button for libraries with zero dependents (Course Presentation's dependency tree makes everything undeletable once both content types are installed together, so the delete test seeds only Blanks), and H5P.MathDisplay isn't a Hub content type (no `h5p.json`, 404s on the content-types API) so its `.h5p` is fetched at runtime from a static h5p.org URL rather than checked into `test/data/`, with the test skipping itself (not failing) if that download doesn't succeed.
- **Session 5** — adds `test/specs/content-lifecycle.spec.ts` covering "Tests: Content" for both H5P.Blanks and H5P.CoursePresentation, parameterised over a shared fixture array: create with metadata (title, license, author), upload an image, fill in minimal content (Course Presentation also adds two extra subcontent elements to its slide), save, play, edit, save again, and delete — verified against both the start page and the play route's response after deletion. Extends `EditorPage` with metadata-author, collapsible-group, image-upload, copy/paste and Course Presentation slide-element helpers, and fixes a pre-existing substring-matching bug in `StartPage.downloadLink()` (it matched both "download" and "download HTML"). Several plan/selector corrections came out of this session: the metadata author widget's real shape, the image-upload flow shared by both content types, the "media type" `<select>` auto-selecting and hiding itself once only one real library option is installed, how newly-added Course Presentation slide elements fully overlap by default (so only the most-recently-added one can be reliably re-selected), and the play route returning HTTP 500 (not 404) for deleted content.
- **Session 6** — adds `test/specs/round-trip.spec.ts` (copy/paste in the editor, and download `.h5p` / re-upload with a structural check via `yauzl-promise`, unzipped in Node rather than trusted through the UI) and `test/specs/html-export.spec.ts` (download the HTML export, load it via a `file://` URL, assert it renders with a clean console), then adds `firefox`, `webkit` and `mobile-safari` (`devices['iPhone 15']`) projects to `playwright.config.ts`. The plan's original cross-browser `testMatch` assumed a `player.spec.ts` file that was never created; the actual matrix instead runs only `html-export.spec.ts` on non-Chromium projects, which seeds its content over the JSON `POST /h5p/new` endpoint instead of the editor UI — driving the CKEditor "Text blocks" field via `.fill()` does not reliably commit its value under Mobile Safari's touch emulation, a genuine editor/WebKit interaction quirk unrelated to what this spec is meant to check. Also extends the shared console-error allowlist with an expected, browser-dependent `contentUserData` 403 (Firefox/WebKit surface it, Chromium doesn't), and fixes an `EditorPage.copyContent()` ambiguity that appears once a field-level copy button is also enabled.
- **Session 7** — adds `test/fixtures/storageEnv.ts` (an `E2E_STORAGE` dimension selecting `fs` | `mongo-only` | `mongo-s3-redis`, transcribing `packages/h5p-examples/*.env` into env-var overlays with `e2e`-prefixed collection/database/bucket names so a run can never collide with `test:h5p-mongos3`'s own fixtures) and a `MongoS3StateResetter` (drops configured Mongo collections, empties/creates S3 buckets, flushes the Redis logical DBs used for cache/lock). Adds root scripts `test:e2e:mongo`, `test:e2e:mongo-s3-redis`, `start:dbs:redis`, `stop:dbs:redis`. Non-fs permutations are restricted to `library-management.spec.ts` + `content-lifecycle.spec.ts` per the plan. Also fixes a real, previously-latent bug this session surfaced: the config never applied session 2's own stated timeout budget ("expect timeout 15s, test timeout 120s"), so a Course Presentation drag-and-drop action reliably timed out under the `mongo-only` permutation (which deliberately runs without a library-metadata cache — see below — and is measurably slower at editor/dragnbar initialization as a result). Applying that budget globally fixed it for every storage mode. Also drops `CACHE` entirely for `mongo-only` (rather than carrying over `mongo+mongos3.env`'s `CACHE=in-memory`), since the example server stays up across the whole permutation run and an in-memory cache has no way to learn that a direct Mongo wipe between spec files changed storage out from under it — `CACHE=redis` doesn't have this problem, so `mongo-s3-redis` keeps Redis-backed caching and the resetter flushes it directly.
  - **Verification note:** `mongo-only` was verified end-to-end against real Mongo + MinIO containers — 9/9 non-network tests passed in a single clean run after the timeout fix, and the fs-mode suite's own reset-across-runs guarantee gives high confidence the same mechanism holds here. `mongo-s3-redis` could **not** be fully verified in the sandbox this session ran in: `docker pull minio/mc` (used only by that compose file's `minio_init` bucket-provisioning service) reliably hung with zero progress over repeated multi-minute attempts, while every other image the suite needs pulled or was already cached without issue — this reads as a registry/rate-limit quirk specific to that one image in that one sandbox, not a defect in the compose file or the code. The `MongoS3StateResetter` class is shared between both permutations (it already creates buckets if missing, specifically because the plain `mongo-s3-docker-compose.yml` used by `mongo-only` has no `minio_init` equivalent), so the untested code path is small and low-risk: the Redis-flush branch, which is a direct, simple `redis` client call. Recommend a maintainer with normal Docker Hub access run `npm run start:dbs:redis && npm run test:e2e:mongo-s3-redis` once to confirm before relying on it in CI.
- **Session 8** — adds `test/specs/content-hub.spec.ts` (`@network`: search and browse the Hub's "Create Content" tab, install and render two content types not pre-packaged locally — H5P.FindTheWords and H5P.Accordion, chosen for needing only a top-level title filled in) and `test/specs/localization.spec.ts` (`?lng=de` against every string live-loaded from `packages/h5p-server/assets/translations/*/de.json` rather than hard-coded, so a translation update can't silently desync the suite: Hub content type names/descriptions, metadata field labels, the metadata popup's author widget, the metadata popup as a modal, the player's Reuse dialog as a modal, a German server error for an invalid `.h5p` upload, and the player's server-rendered "Reuse" button label). Extends `EditorPage` with `searchHub()`/`hubTile()`/`installContentTypeFromHub()`, and fixes `openMetadata()` (previously matched the English-only "Metadata" button text; now uses a CSS selector on the toggle wrapper so it works under any language). The `@offline` fixture-recording variant the plan floated for the Hub spec was deliberately not built — installs are already fast (under 2s), and hand-maintained fixtures for a Hub UI that isn't code in this repo looked like more ongoing cost than it would save.
- **Session 9** — adds a `rest-example` Playwright project smoke-testing `packages/h5p-rest-example-server` + `packages/h5p-rest-example-client` (log in as a teacher, create Fill in the Blanks content through the `<h5p-editor>`/`<h5p-player>` web components, play it, delete it), and a `RestExampleAppPage` page object. This app has its own CSRF protection and admin-only library-upload permission (unlike `h5p-examples`), so seeding goes through a dedicated `seedRestExampleLibrary()` fixture that logs in as `admin` first. Also fixes two pre-existing issues surfaced while wiring this up: the Vite client's dev-server proxy hardcoded port 8080 (now configurable via `H5P_REST_SERVER_PORT`, needed since the REST server runs on 8081 to avoid colliding with `h5p-examples`), and a real dependency mismatch (the client pinned `vite@7.3.6` while `@vitejs/plugin-react` resolved a separately-hoisted `vite@8.2.2` at runtime, breaking every module transform with `Missing field 'moduleType'` even on a plain `npm run start:rest:client` — fixed by aligning the client's own `vite` devDependency to `8.2.2`). Also documents a real DOM difference from the server-rendered player: the `<h5p-player>` web component always wraps content in a real `iframe.h5p-iframe`, unlike `packages/h5p-examples`' player page.
- **Session 10** — adds an `e2e-tests` job to `.github/workflows/ci.yml` (`needs: install-build`) running the default Chromium/filesystem project with `@network` excluded on every push, with Playwright browser caching and `playwright-report`/`test-results` uploaded via `actions/upload-artifact@v4` with `if: always()`; added to the `release` job's `needs:` alongside every other quality gate, left out of `coveralls-finish` since it produces no coverage. Adds `.github/workflows/e2e-nightly.yml`, a scheduled (+ `workflow_dispatch`) workflow for the expensive dimensions that shouldn't run on every push: the full browser matrix, the REST example project, both storage permutations, and the `@network` Hub tests — it builds its own workspace artifact rather than sharing `ci.yml`'s, since GitHub Actions artifacts can't be shared across separate workflow runs. Rewrites `test-plan.md`: automated coverage is now a table pointing at the spec file covering each area, and the residual manual checklist (Lumi, WordPress, real Mobile Safari, a visual sanity pass, the `test:h5p-redis-lock` unit test) is down to **5 checkboxes**, well under the ~20 target.

## The whole plan is now complete
`E2E_AUTOMATION_PLAN.md` is marked complete at the top of the document. `test-plan.md` has been slimmed from ~30 checkboxes across many sections down to a 5-item residual manual checklist plus a coverage table and local-run instructions.

**Important caveat on session 10's CI changes:** this session's environment does not have a GitHub token with `workflow` OAuth scope, so the commit touching `.github/workflows/ci.yml` and adding `.github/workflows/e2e-nightly.yml` could **not be pushed** (GitHub rejects it server-side: *"refusing to allow an OAuth App to create or update workflow `.github/workflows/ci.yml` without `workflow` scope"* — confirmed both via `git push` and the Contents API, so it isn't a git-vs-API quirk, it's the token itself). Every other session-9/10 change (the REST example spec, `test-plan.md`, `E2E_AUTOMATION_PLAN.md`) is pushed and part of this PR. **The workflow commit exists locally on this session's checkout of `sr258/automate-test-plan` but is not yet on the remote branch or in this PR's diff.** To land it: run `gh auth refresh -s workflow` (or otherwise obtain a token with `workflow` scope) and `git push`, or apply `.github/workflows/ci.yml`'s diff and `.github/workflows/e2e-nightly.yml` manually — see this PR's session 10 commit message (once pushed) or `E2E_AUTOMATION_PLAN.md`'s session 10 section for the exact content and reasoning.

## Test plan
- [x] `npm run build && npm run test:e2e` passes from a clean checkout
- [x] `npm run test:e2e` (default Chromium project, fs storage) passes 3 consecutive times, including from a wiped `packages/h5p-examples/h5p/{content,libraries}` state with no manual steps
- [x] Each spec file passes in isolation and as part of the full run; the full default run (21 tests, `@network` excluded) completes in well under the ~10 minute target (~30s locally)
- [x] `npx playwright test --project=webkit` passes
- [x] `npm run test:e2e:mongo` (the `mongo-only` storage permutation) passes 9/9 against real Mongo + MinIO containers
- [ ] `npm run test:e2e:mongo-s3-redis` (the `mongo-s3-redis` permutation) — implemented but **not verified** in this session's sandbox due to an environment-specific Docker image pull that would not complete; see session 7's note above
- [x] `npx playwright test --project=rest-example` passes 3 consecutive times (session 9)
- [x] `npm run lint` and `npm run format:check` are clean
- [x] `npx tsc --noEmit` clean for `packages/h5p-e2e`
- [x] `npm test` (full unit suite, 360 tests, via pre-push hook) passes
- [ ] **CI-only-by-inspection, not run in this sandbox:** the `e2e-tests` job in `ci.yml` and the whole `e2e-nightly.yml` workflow. Validated by YAML parsing (`python3 -c "import yaml; yaml.safe_load(...)"`) and close comparison against `ci.yml`'s existing, already-battle-tested patterns (artifact download/extract, `db-tests`' manual MinIO startup, `actions/cache` keyed on a tool version) — not by a real GitHub Actions run, and (per the caveat above) not yet even pushed to this branch.

https://claude.ai/code/session_01BNXYn7YhatBFp8cVCDbCVD


sr258:
## Second review round — 7 findings fixed (a53019c7)

A follow-up review of the E2E suite turned up seven issues; all are fixed in `a53019c7`. Grouped by kind:

### Races / silent failures

- **`EditorPage.installContentTypeFromHub`** — `installButton.isVisible()` was evaluated immediately after the tile click and does not auto-wait. The Hub renders its detail panel asynchronously, so on a slow run the install was skipped entirely and the method hung on `.field-name-extraTitle input` until the 120s test timeout, blaming the title field. The branch was dead in any case: `content-hub.spec.ts` resets library storage in `beforeAll` and seeds neither type it uses, so the detail panel is the only possible outcome. The method is now linear and fails at the install button, naming what actually broke.
- **`reset.ts` `resetRedis()`** — always connected to `REDIS_HOST`/`REDIS_PORT`, including for the DB derived from `LOCK_REDIS_DB`, although `createH5PEditor.ts` treats `LOCK_REDIS_HOST`/`LOCK_REDIS_PORT` as independent settings. Any split configuration flushed the wrong instance, and `LOCK=redis` without `CACHE=redis` built a client with `host: undefined, port: NaN`. It only worked because `MONGO_S3_REDIS_ENV` points both at `localhost:6379`.
- **`reset.ts` `emptyBucket()`** — the S3 `DeleteObjects` response was discarded, but that API reports per-key failures in an `Errors` array on an HTTP 200. Partial failures let `reset()` claim success while stale content files survived into the next spec. Now throws, naming the failed keys.

### Vacuous assertions

- **"lists no libraries right after a reset"** — passed before the admin panel mounted: `toBeHidden()` passed because the element was not attached, and `toHaveCount(0)` because no `tbody` existed. Now asserts the loaded state positively first, then the row count.
- **Content type cache update** — `before` could be snapshotted while the label still read `Loading...`, so the later `.not.toBe(before)` was satisfied by loading finishing rather than by the update succeeding (`updateCache()` swallows its errors). Now waits past `Loading...` before capturing the baseline.

### Config

- **`playwright.config.ts` base URL** — the `h5p/core` guard and the `webServer` bypass both keyed on `E2E_BASE_URL`, but the rest-example project's URL comes from `E2E_REST_EXAMPLE_BASE_URL`. Pointing that run at already-running servers still spawned both `webServer` entries, colliding on ports 3000/8082 under CI (`reuseExistingServer: false`).
- **`playwright.config.ts` storage restriction** — Playwright replaces rather than merges `testMatch`, so the top-level pattern limiting `mongo-only`/`mongo-s3-redis` runs to `library-management` + `content-lifecycle` was escaped by `firefox`, `webkit`, `mobile-safari` and `rest-example`, each of which defines its own. `E2E_STORAGE=mongo-only` without `--project` would have run `rest-example-smoke` against a server with no Mongo config; it worked only because every script passes `--project=chromium`. The restriction is now folded into each project's pattern; `chromium`'s behaviour is unchanged.

Also checked and cleared, no change needed: `test.beforeAll({ request })` is legal in the installed Playwright; the German `chooseContentType('Fill in the Blanks')` substring match holds against `hub/de.json`; `FsStateResetter` directory names and all `storageEnv.ts` variables match the example apps; and the `package-lock.json` churn is a benign dedupe (drops the nested `vite@7.3.6` under `h5p-rest-example-client`, root was already on 8.2.2, within `@vitejs/plugin-react`'s peer range).

Verification: `tsc --noEmit`, `npm run lint`, `npm run format:check` and the full unit suite (360 tests, 39 files) all pass. The E2E suite itself was not run locally — it needs the servers and Docker, so CI covers it.

PR #4625, opened 2026-09-13
fix(h5p-server): reject packages with unsatisfied library dependencies

sr258:
## Problem

Importing "Arts of Europe - Branching scenario" from the Content Hub looked like a silent failure: a green "successfully imported" banner, then a blank editor.

The import actually succeeded. The request right after it did not:

```
GET /h5p/ajax?action=libraries&machineName=H5P.BranchingScenario&majorVersion=1&minorVersion=4
→ 404 {"errorCode":"","httpStatusCode":404,
       "message":"The library H5P.Question-1.5 is not installed on this system.","success":false}
```

`h5peditor.js` only logs the status text of a failed AJAX request to the browser console, so the precise message never reaches the user.

The package is at fault: it ships `H5P.Summary 1.10` and `H5P.SingleChoiceSet 1.11`, both declaring `preloadedDependencies: H5P.Question 1.5`, but bundles only `H5P.Question 1.4.10`. `PackageValidator` never checked this — `PackageImporter` only verified the dependencies declared in the content's own `h5p.json`.

## Change

`PackageValidator` gets a `libraryDependenciesMustBeSatisfied` rule (applied when libraries are installed): it collects the `preloaded`, `editor` and `dynamic` dependencies of every library in the package and reports each one that is neither shipped in the package nor already installed as `missing-required-library` — the same error code the H5P PHP core uses (`h5p.classes.php:1078`).

That code was chosen deliberately: the H5P hub client discards server error messages unless the response carries a `details` array containing an item with exactly that code, in which case it renders each item's message as a list. `expressErrorHandler` already produces that array for `AggregateH5pError`, so the user now sees this with **no change to the externally maintained client**:

```
Download failed
You are not authorized to install or update the libraries required by this content.
  • Missing required library H5P.Question-1.5
  • Missing required library H5PEditor.ColorSelector-1.2
```

The intro sentence is the hub client's own hardcoded text for that code and is misleading here; it can't be changed server-side, and any other code loses the list entirely.

The content's own `h5p.json` dependencies (as opposed to dependencies libraries declare among themselves) are checked separately, in `PackageImporter`. That path used to throw a plain `H5pError('install-missing-libraries')`, which `expressErrorHandler` cannot turn into the `details` array the hub client needs — the same silent failure, on the other half of the import path. It now throws an `AggregateH5pError` with one `missing-main-library` or `missing-required-library` entry per missing library (mirroring the PHP core's `h5p.classes.php:1074-1078`), alongside an `install-missing-libraries` entry kept for downstream code that matches on that code.

### Opt-out

Packages that imported successfully before this change now fail at import, with no way to opt out. `IH5PConfig` gains a required `validateLibraryDependencies` member (default `true`, user-configurable like `enableLrsContentTypes`) that gates the whole check. Integrators who only import content to view or export it — where a blank editor isn't a concern — can set it to `false` to restore the previous behaviour without forking. See `docs/development/decisions/0001-library-dependency-validation.md` and `docs/advanced/security.md` for the reasoning.

### Performance

`LibraryManager.getNotInstalledLibraries` used to call `listInstalledLibraries()`, which issues one `getLibrary()` (a `library.json` read) per installed library purely to populate fields the method never used. It now compares against `getInstalledLibraryNames()` instead — this method now runs twice per package import (once from the new validator rule, once from `PackageImporter`), so on a system with several hundred libraries this removes several hundred redundant storage round trips (network calls on `h5p-mongos3`/S3) per import.

`PackageValidator` also no longer parses every `library.json` twice: `librariesMustBeValid` now collects the parsed metadata of every library it validates into a map that `libraryDependenciesMustBeSatisfied` consumes, instead of re-reading and re-parsing each `library.json` itself.

## Consequences — please review before merging

The check refuses more real Hub content than expected. In each case I verified the package is genuinely broken today, not a false positive — installed into an empty system, then `getLibraryData` for its main library:

| Package | Unsatisfied dependency | Behaviour before this change |
| --- | --- | --- |
| Arts of Europe (hub content) | `H5P.Question-1.5`, `H5PEditor.ColorSelector-1.2` | imports, then editor 404s |
| Making a strawberry smoothie (hub content) | `H5PEditor.ColorSelector-1.2` | installs 54 libraries, then `library-missing` |
| `H5P.BranchingScenario.h5p` (content type) | `H5P.InteractiveVideo-1.27` | installs 70 libraries, then `library-missing` |

These packages ship a different version of a shared editor library than their own older libraries require (`ColorSelector 1.3` shipped / `1.2` required, `Question 1.4` shipped / `1.5` required). On h5p.org this goes unnoticed because those sites have every version installed side by side.

So `H5P.BranchingScenario` can no longer be installed from the content type hub on a fresh system, and a share of Content Hub content is refused. Both were already unusable in the editor — what changes is that the user is told why (and, per above, can opt out of the whole check if that trade-off doesn't fit their deployment). The ADR discusses the alternatives (report-but-import; lenient same-major/higher-minor resolution; surfacing the message in the editor instead) and why they were not chosen.

Also note the check is **permission-gated**: `PackageImporter` only runs it when the acting user has `GeneralPermission.UpdateAndInstallLibraries`. For users without that permission, the rule never runs and the blank-editor failure mode is still reachable for them.

## BREAKING CHANGE

- Packages whose libraries declare unsatisfiable dependencies are now refused at import instead of installing and failing later in the browser.
- `H5P.BranchingScenario` is no longer installable from the content-type hub on a fresh system.
- `IH5PConfig` gains a required `validateLibraryDependencies` member — a breaking change for anyone implementing `IH5PConfig` directly rather than using the bundled `H5PConfig`.

## Docs

`docs/development/decisions/0001-library-dependency-validation.md` (first ADR in the repo, linked from `docs/SUMMARY.md`) is **accepted**, and documents the `validateLibraryDependencies` opt-out, the permission gating, why `dynamicDependencies` are included, and the content-level fix. The option is also documented in `docs/advanced/security.md` next to the existing library-restriction guidance.

## Tests

- `PackageValidator` unit tests cover `preloadedDependencies`, `editorDependencies` and `dynamicDependencies` declarations, a dependency satisfied by an already-installed library, an unparseable `library.json` (reported once, not twice), and the config gate turned off.
- `PackageImporter` tests cover both content-level shapes: an uninstalled ordinary dependency and a missing main library.
- `LibraryManager` tests cover same-major.minor-different-patch (counts as installed), different-minor (counts as missing), and spy-assert that `getLibrary` is no longer called.
- `test/helpers/installPackages.ts`: the integration tests that install all Hub packages into one system relied on later packages providing earlier ones' dependencies, so failed packages are retried after the others; when no more progress is possible, every still-failing package is now named in one aggregated error instead of only the last one processed.
- The export round-trip and html-exporter tests skip `H5P.BranchingScenario.h5p`, which can no longer be imported standalone; the shared list of such packages now lives in one module (`test/data/hub-content.ts`) instead of two copies.
- `Validation.test.ts`'s mock reports all dependencies as installed — that test checks package structure, not host inventory.

`npm run build`, `npm run lint`, `npm run format:check`, `npm test` (372) and `npm run lint:docs` all pass.

https://claude.ai/code/session_013YC1AE9hDGJrcFuUbbhB1G

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -39,4 +39,4 @@
   - [Core updates](development/core-updates.md)
   - [Project Status](development/status.md)
   - Decisions
-    - [Validating library dependencies when importing packages](development/decisions/0001-library-dependency-validation.md)
+    - [Handling packages whose library dependencies can't be resolved](development/decisions/0001-library-dependency-validation.md)
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -38,3 +38,5 @@
   - [Testing & code quality](development/testing-quality.md)
   - [Core updates](development/core-updates.md)
   - [Project Status](development/status.md)
+  - Decisions
+    - [Validating library dependencies when importing packages](development/decisions/0001-library-dependency-validation.md)
--- a/docs/development/decisions/0001-library-dependency-validation.md
+++ b/docs/development/decisions/0001-library-dependency-validation.md
@@ -1,10 +1,10 @@
 ---
-title: Validating library dependencies when importing packages
+title: Handling packages whose library dependencies can't be resolved
 group: Documents
 category: Contributing
 ---
 
-# 1. Validating library dependencies when importing packages
+# 1. Handling packages whose library dependencies can't be resolved
 
 Date: 2026-09-13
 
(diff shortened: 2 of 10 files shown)


sr258:
## Review follow-ups — non-blocking, for later pickup

The one blocking finding from review (the shared hub-content fixture module
resolving its path against the current working directory, which broke
`npm test` inside `packages/h5p-html-exporter`) is fixed in a183a061.

The items below are deliberately **not** fixed in this PR. None of them
changes behaviour; they are clarity, typing and tooling-coverage gaps. Listing
them here so they aren't lost.

### 1. `getNotInstalledLibraries` silently dropped patch-version matching

`LibraryManager.ts:303`. The previous implementation compared via
`InstalledLibrary.compareVersions`, which *does* take `patchVersion` into
account when the passed object carries one. The new comparison ignores patch
versions unconditionally.

This is not a bug today: both call sites (`PackageImporter.ts:327`,
`PackageValidator.ts:697`) pass dependency objects parsed from `h5p.json` /
`library.json`, which per the H5P spec never carry a patch version, so the old
code degenerated to major.minor equality too. But the JSDoc doesn't say so, and
a future caller passing an `IInstalledLibrary` would get quietly different
matching than before. Worth a sentence stating that matching is by
`machineName` plus exact major.minor.

### 2. `getNotInstalledLibraries` is now O(installed × requested)

`LibraryManager.ts:308`. The old code looked up by `machineName` key first; the
new one runs a linear `.some()` over every installed library for each requested
one. Irrelevant at realistic library counts, and still far cheaper than the
per-library `getLibrary()` reads this commit removed — but building a `Set` of
uberNames once would be both faster and shorter than the current three-way
comparison.

### 3. `Map<string, any>` for the library metadata map

`PackageValidator.ts:164` and the signatures of
`libraryDependenciesMustBeSatisfied`, `librariesMustBeValid`,
`collectLibraryMetadata` and `validateLibrary`. `any` is permitted in this
codebase but discouraged, and the values here are parsed `library.json`
content, for which `ILibraryMetadata` already exists. Typing the map would also
document what the consuming rule is allowed to rely on.

### 4. `ContentMetadata.toUbername` can return `undefined`

`PackageImporter.ts:355`. If `mainLibrary` is not among the content's
`preloadedDependencies`, `toUbername` returns `undefined`; the comparison then
never matches and every missing library is reported as
`missing-required-library` rather than `missing-main-library`. That degrades
gracefully and is arguably correct, so no change is needed — but the comment
at that spot currently explains only the hyphen/whitespace separator hazard,
and this second case is worth a clause.

### 5. `test/data/hub-content.ts` is outside every tooling glob

The new shared fixture module sits at the repository root, where:

- the root `tsconfig.json` `include` list covers only `packages/*/{src,test}`,
  so it is pulled in transitively but never type-checked on its own;
- `format:check` globs `packages/*/{src,test,examples}/**/*.{ts,tsx}`, so
  prettier never checks it.

Adding `test/**/*` to both would close the gap. This is how the cwd-dependent
path in it went unnoticed.

### 6. Sanity-check the release labelling after a183a061

The `BREAKING CHANGE:` footer lives in eb3a1a3e, which is no longer the last
commit on the branch. `lerna version --conventional-commits` analyses every
commit in the release range rather than only the tip, so this should be fine —
but since the labelling is the thing that decides major vs. patch here, it is
worth confirming on the release branch rather than assuming.

### 7. Evidence gap: no "fails without the fix" check for the content-level half

Verification disabled the `addRuleWhen` registration at
`PackageValidator.ts:235` and confirmed exactly the three dependency-rejection
tests fail, so the validator half is demonstrably load-bearing. There is no
equivalent demonstration for the `PackageImporter` half: its two new tests
assert the `AggregateH5pError` shape, which only exists on this branch, so they
can't be run against `master` to watch them fail. Coverage is fine; only the
evidence is one-sided.

